PR #7, opened 2015-02-07
Feature/repos option

jhoblitt:


jhoblitt:
This is dependent on https://github.com/LSST/lsst_build/pull/1

etc/settings.cfg.sh
```diff
@@ -25,9 +25,7 @@ export EUPSPKG_SOURCE=${EUPSPKG_SOURCE:-git}
 export EUPS_PKGROOT=$LSSTSW/distserver/production
 
 # the location of source repositories
-BASE='git://git.lsstcorp.org/LSST'
-SBASE='https://stash.lsstcorp.org/scm'
-export REPOSITORY_PATTERN="$BASE/sims/%(product)s.git|$BASE/DMS/%(product)s.git|$BASE/DMS/devenv/%(product)s.git|$BASE/DMS/testdata/%(product)s.git|$BASE/external/%(product)s.git|$SBASE/sim/%(product)s.git"
+export REPOFILE=${LSSTSW}/etc/repos.yaml
```
ktlim: Uh oh.  `--repos="${REPOSFILE}"` above; `export REPOFILE` here.
Also, I think it's a good idea to still have a backup REPOSITORY_PATTERN.  Furthermore, it might be good to have the `repos.yaml` only contain things that need to be remapped, rather than everything.

jhoblitt: Nice catch.  I will repush in a moment.

What's the value of having a REPOSITORY_PATTERN if the location of every repository is explicitly mapped?  That could both hide errors in `repos.yaml` and be non-deterministic.  There's also some logic that may prevent a repos.yaml supplied url from changing the remote in an existing clone if its remote matches a patterns generated url.

I left the pattern functionality in because I'm unsure of how this package is being used, if at all, outside of lsst-dev.  I'd be happy to remove the functionality completely or add the ability to fetch a repos.yaml file via HTTP so any existing users wouldn't be required to supply a configuration file.  However, for our live build system we need to achieve two things:

1) Fully decouple the repository name from the EUPS package name.  We've already demonstrated this by renaming the github `pex_harness` repo to `legacy-pex_harness`.

2) Have deterministic git remotes; as it is now, a repository being created at the correct path could automatically supersede the desired repo (or not, untli the existing clone is deleted).

ktlim: The point is *not* to have every repository explicitly mapped -- only those that can't fit the pattern  After we move, I'm guessing we'll have only a couple of entries in the REPOSITORY_PATTERN.  It's much easier administratively not to have to update a file (and all users of that file) every time a product is added.  Since we'll only have a couple of potential locations, and since it will be a huge problem to have duplicate names anyway, I don't think the "creating a new repository at a preceding path" is really much of a problem.

frossie: I don't like this. "We'll only have a couple of potential locations" is speculation, and duplicate names would actually be undetected under the current regime. Having two mechanisms for something that is only relevant to our build system is a completely unnecessary increase in maintainance workflow. I suggest we talk about this further in the repos reorg session on Tuesday.

PR #12, opened 2015-03-21
(maint) change $LSSTSW to default to $HOME/lsstsw

jhoblitt:
We should not assume that $LSSTSW == $HOME as this is an atypical
configuration.

### DIFF
--- a/etc/settings.cfg.sh
+++ b/etc/settings.cfg.sh
@@ -16,7 +16,7 @@ fi
 #
 
 # where are we? default to $HOME if $LSSTSW hasn't been defined
-export LSSTSW=${LSSTSW:-$HOME}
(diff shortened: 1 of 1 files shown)


etc/settings.cfg.sh
```diff
@@ -16,7 +16,7 @@ fi
 #
 
 # where are we? default to $HOME if $LSSTSW hasn't been defined
-export LSSTSW=${LSSTSW:-$HOME}
+export LSSTSW=${LSSTSW:-$HOME/LSSTSW}
```
mjuric: You probably want lowercase lsstsw here (i.e., "$HOME/lsstsw")?

jhoblitt: gez -- I'm going home after this. :)

jhoblitt:
Repushed

PR #14, opened 2015-03-25
Changing location of throughputs repo.

mareuter:


jhoblitt:
@mareuter Thanks!  This change is now live on lsst-dev.

```
$ grep throughputs /home/lsstsw/etc/repos.yaml 
throughputs: https://github.com/lsst/throughputs.git
```

PR #18, opened 2015-03-30
(DM-2371) add -n flag to rebuild script

jhoblitt:
If specified, causes the --no-fetch flag to be passed to `lsst_build prepare`.
This may be useful for testing.


mjuric:
+1, merge it.

PR #20, opened 2015-04-10
bin/deploy: update the conda system package to 5.8-2 or later

mjuric:
This version isn't affected by the DM-1801 bug (does not ship with its own libm).

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -34,6 +34,17 @@ test -f "$LSSTSW/anaconda/.deployed" || ( # Anaconda
     echo "::: Deploying Anaconda ${ANACONDA_VERSION} for ${ana_platform}"
     curl -# -L -O http://repo.continuum.io/archive/${ana_file_name}
     bash ${ana_file_name} -b -p "$LSSTSW/anaconda"
+
+    # workaround for libm issue (DM-1801); remove once Anaconda version 
+    # is bumped above 2.1.0
+    if [[ $ANACONDA_VERSION == 2.1.0 ]]; then
+    	PATH="$LSSTSW/anaconda/bin:$PATH" conda update --yes system
+    else
+        # Make sure we don't forget to remove this
+        echo "You've bumped the anaconda version above 2.1.0, now remove this workaround!"
+        exit -1
+    fi
+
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -34,6 +34,17 @@ test -f "$LSSTSW/anaconda/.deployed" || ( # Anaconda
     echo "::: Deploying Anaconda ${ANACONDA_VERSION} for ${ana_platform}"
     curl -# -L -O http://repo.continuum.io/archive/${ana_file_name}
     bash ${ana_file_name} -b -p "$LSSTSW/anaconda"
+
```
jhoblitt: The value for `ANACONDA_VERSION` presumably should be updated.

bin/deploy
```diff
@@ -34,6 +34,17 @@ test -f "$LSSTSW/anaconda/.deployed" || ( # Anaconda
...
+    # workaround for libm issue (DM-1801); remove once Anaconda version 
+    # is bumped above 2.1.0
+    if [[ $ANACONDA_VERSION == 2.1.0 ]]; then
+    	PATH="$LSSTSW/anaconda/bin:$PATH" conda update --yes system
+    else
+        # Make sure we don't forget to remove this
+        echo "You've bumped the anaconda version above 2.1.0, now remove this workaround!"
+        exit -1
```
jhoblitt: Does it make sense to exit here?  This would only have affect for existing lsstsw installation re-running `deploy` and will fail for new deployments once `ANACONDA_VERSION` is updated.

bin/deploy
mjuric: No -- we want to keep it to 2.1.0 (2.2.0 forces an upgrade to an incompatible version of IPython -- which users will hate), and just update the 'system' conda package (that itself resolves the libm bug).

bin/deploy
mjuric: That's the idea -- make this unusable unless this whole block is deleted.

This is a (somewhat harsh) reminder for developers, really -- once ANACONDA_VERSION is changed from 2.1.0, we should stop doing the 'conda update --yes system' hack.

PR #24, opened 2015-05-26
lsstsw was not deploying on Macs

danielsf:
The Mac distribution of Anaconda does not contain the conda 'system' package.  Thus, when we try to update 'system' while installing anaconda, we get an error.  This was fixed in the anaconda package proper.  The fix did not make it to the lsstsw deploy script.  This pull request copies the fix from the anaconda eupspkg.cfg.sh script to lsstsw/bin/deploy.


danielsf:
It looks like there may be other problems beyond not deploying on Macs.  I'm going to close this pull request and re-issue it after I can successfully build a package using lsstsw.

PR #25, opened 2015-05-26
lsstsw was not deploying on Macs

danielsf:
Sorry.  User error.  This is ready for review.

To recap: Macs do not contain the 'system' package, thus, updating 'system' when we install anaconda causes an error.  I just copied the workaround from anaconda/ups/eupspkg.cfg.sh to the lssts/bin/deploy script.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -38,7 +38,11 @@ test -f "$LSSTSW/anaconda/.deployed" || ( # Anaconda
     # workaround for libm issue (DM-1801); remove once Anaconda version 
     # is bumped above 2.1.0
     if [[ $ANACONDA_VERSION == 2.1.0 ]]; then
-    	PATH="$LSSTSW/anaconda/bin:$PATH" conda update --yes system
+        if ! PATH="$LSSTSW/bin:$PATH" conda list --no-pip --json system >/dev/null; then
+            echo "No system package to upgrade"
+        else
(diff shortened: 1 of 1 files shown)


jhoblitt:
@danielsf Is this also the case with anaconda 2.2.0?

danielsf:
Based on the discussion on JIRA issue DM-1801, we are performing the following steps when we deploy the anaconda eups package

1) download and install anaconda
2) use pip to update the 'system' package to get rid of a bug in anaconda's libm.so

step (2) causes an error on Macs because the 'system' package does not exist for OSX.  It sounds like, if we go to anaconda 2.2.0, this whole process becomes unnecessary and the problem will go away.

So... I guess the answer to your question is "no; but going to anaconda 2.2.0 will require changes to the deploy script, anyway"

jhoblitt:
I'm OK with merging this but I'm wondering if it wouldn't be better to upgrade to 2.2.0.  @mjuric had concerns about ipython compatibility https://github.com/lsst/lsstsw/pull/20#discussion_r28176893 but I believe some folks have been using 2.2.0

jhoblitt:
Per discussion on hipchat, an RFC to upgrade anaconda to 2.2.0 has been opened: https://jira.lsstcorp.org/browse/RFC-54

jhoblitt:
This should be resolved by #27 .

PR #27, opened 2015-06-08
update anaconda version to 2.2.0

jhoblitt:
Per RFC-54. Also resolves the OSX build issue reported in PR #25.


jhoblitt:
I've manually tested that ./deploy succeed and does install anaconda 2.2.0 on a el6 VM.

PR #34, opened 2015-07-15
(DM-3151) repos-lint

jhoblitt:


bin/repos-lint
```diff
@@ -0,0 +1,88 @@
...
+        validate_product_name(pkg)
+    except ValueError as e:
+        print(e)
+        return False
+
+    print('OK')
+    return True
+
+def validate_string(s):
+    if not type(s) is str:
```
timj: This should use `isinstance`

bin/repos-lint
```diff
@@ -0,0 +1,88 @@
...
+repos=0
+problemos=0
+
+print('repos')
+print('===')
+print()
+
+with open(args.filename, 'r') as f:
+    data = yaml.safe_load(f)
+    if not type(data) is dict:
```
timj: I guess it would be more normal to ask if this was an instance of dict.

bin/repos-lint
jdswinbank: Is this necessary at all? Why insist that `s` is a string, rather than just a "thing that `urlparse` can handle"?

bin/repos-lint
```diff
@@ -0,0 +1,88 @@
...
+        print('invalid format: must be a hash')
```
timj: Here be Perl Speak.

bin/repos-lint
```diff
@@ -0,0 +1,88 @@
...
+        if not validate_product(k):
+            problemos += 1
+        if not validate_repo(v):
+            problemos += 1
+
+print()
+print('summary')
+print('===')
+print('%16s: %d' % ('total repos', repos))
+print('%16s: %d' % ('errors', problemos))
```
timj: So you can have more errors than you have repos? Maybe keep track of the naming errors differently from repo errors?

bin/repos-lint
jhoblitt: The intent is to catch formatting errors, such as indented lines that yaml will parse as a nested hash, with an intelligible error message.

timj: I guess the check here is that the YAML hasn't returned a number or a weird object. This function is called to check the package name and the URL.

bin/repos-lint
```diff
@@ -0,0 +1,88 @@
...
+    print('%16s: ' % 'repo url', end='')
+
+    try:
+        validate_string(repo)
+        url = urlparse(repo)
+
+        if url.scheme == 'https':
+            urlopen(repo)
+        elif url.scheme == 'git':
+            # don't want to attempt to clone the repo as it might be large, so
```
timj: In theory you could use the github API to check the repo exists (if it's a Github repo in the first place). Probably not worth the effort.

bin/repos-lint
jhoblitt: Actually, it's YAML speak.  Technically, it's a mapping but the common usage seems to be hash.

bin/repos-lint
jhoblitt: I believe the usage of `type()` is correct here because we're imposing limits on what should be contain in the yaml file.  We don't want to allow some sort of serialized object that's a subclass (because that really shouldn't be in the configuration file...).

bin/repos-lint
timj: Ok. It looks weird checking for a dict and reporting it's not a hash. Can the error message be clearer though. How about:
```
YAML file has an invalid format: Must be a hash.
```
At least that would show people that the error refers to the YAML file explicitly.

bin/repos-lint
jhoblitt: Yes, at the moment there are 2 checks per "repo" but there could be more in the future.  I started out only intending to do basic "does it parse" validation but the scope creeped as I experimented with inserting errors.  It would probably make sense to eventually convert it to use a real test harness.

bin/repos-lint
timj: Please add a short comment then indicating that you are being deliberately restrictive.

bin/repos-lint
jhoblitt: I thought about it -- it might be a bit faster / save some bandwidth but unauthenticated API queries are rate limited to 60/hour.  Forcing folks to provide credentials for an occasional sanity checker seemed excessive.

timj:
Looks good now. Ok to merge.

jhoblitt:
Thanks for the review!

PR #38, opened 2015-07-28
Add daf_ingest to repos.yaml

smonkewitz:


jhoblitt:
travis passed while a test jenkins setup (which was working) failed because the container is having trouble downloading .deb files for some reason today.  This is good to merge.

PR #39, opened 2015-07-28
(DM-3236) Deleting a remote branch is not noticed by Jenkins

jhoblitt:


jhoblitt:
@mjuric I'm doubtful that this will fix the problem @r-owen reported but I needed to make these changes for testing so my logic is we might as well merge them.

jhoblitt:
(lsst_build made it through prepare for the entire stack with 2.2.2 -- a full build is still running)

mjuric:
OK to merge, though I agree it may be that the fetch line (which I did to speed up the fetch) may need improving, rather than this being a bug in git.

Issue #40, opened 2015-07-31
curl installed by anaconda fails to fetch https URLs during later deploy steps


jrk:
Following the directions from https://confluence.lsstcorp.org/display/LDMDG/The+LSST+Software+Build+Tool on a fairly vanilla Ubuntu 14.04/x86_64 machine, `deploy` completes its first major step of installing anaconda then immediately fails trying to fetch the git source with a variant of this error:

```
curl: (77) error setting certificate verify locations:
```

It seems this occurs because (1) `anaconda/bin` is prepended to the `$PATH`, and (2) anaconda includes its own `anaconda/bin/curl` which then takes over subsequent `curl` invocations in the deploy script. This seems to be misconfigured by default for SSL, causing the error (apparently looking for CA data in a nonexistent path).

The obvious workaround (without disabling SSL certificate validation) is simply to force use of the system `curl` by capturing its path at script startup. stanford-gfx/lsstsw@d95aae6b22 hacks this in for common configurations by hard-coding use of `/usr/bin/curl`.

jhoblitt:
@jrk Thanks for reporting this.  We probably can't hardcode the curl path because of OSX. We might be able to change the `PATH` ordering so that anaconda binaries don't override what's supplied by the system.

jrk:
Hard-coding the path definitely seems wrong (why I didn't send a pull request yet), though OS X does also use /usr/bin/curl as its system default.

That said, I think you could potentially do something like `CURL=$(which curl)` at the start, before the `PATH` is extended.

jhoblitt:
I'm not sure if anaconda needs to be in the search path at all.

mjuric:
@jrk @jhoblitt I'll open an issue with the Continuum guys.

In the meantime, I think you can add `conda remove curl` right after Anaconda gets installed (it's not needed, since the system already has it).

mjuric:
Btw., I have a PR almost ready to replace Anaconda with Miniconda (I think it doesn't pull in `curl` by default).

mjuric:
FYI, this is the upstream issue -- https://github.com/conda/conda-recipes/issues/352..

timj:
Can this issue be closed now that we use miniconda?

Also, that upstream issue is no longer accessible. Did all the issues associated with conda-recipes get deleted???

jrk:
I haven't been using this for a while, so for my sake, at least: certainly.

PR #41, opened 2015-07-31
Switch to Miniconda instead of Anaconda

mjuric:
Reasons:
 * Reduces the size of the download by 50-75%
 * Allows tight control of dependencies (i.e., only necessary python
   packages are installed, making it easy to detect if a stack package
   begins using an unapproved dependency).

Caveats:
 * Should be verified it works by rebuilding the whole stack first (I only built `lsst_distrib` on a Mac).

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -14,7 +14,6 @@ set -e
 
 mkdir -p ${LSSTSW}/{sources,build,var/run,var/log,lfs,distserver/production}
 
-export PATH="$LSSTSW/miniconda/bin:$PATH"
 export PATH="$LSSTSW/lfs/bin:$PATH"
 export PATH="$LSSTSW/bin:$PATH"
 
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,45 +7,55 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
 source ${SCRIPT_DIR}/../etc/settings.cfg.sh
 
 EUPS_VERSION=${EUPS_VERSION:-1.5.9}         # Version of EUPS to install
-ANACONDA_VERSION=${ANACONDA_VERSION:-2.2.0} # Version of Anaconda to install
+MINICONDA_VERSION=${MINICONDA_VERSION:-latest} # Version of Miniconda to install
 GIT_VERSION=${GIT_VERSION:-2.2.2}           # Version of git to install
 
 set -e
 
 mkdir -p ${LSSTSW}/{sources,build,var/run,var/log,lfs,distserver/production}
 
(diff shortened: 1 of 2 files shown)


bin/deploy
```diff
@@ -7,45 +7,55 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
...
                 install_name_tool -id $entry $entry
             done
     fi
+
+    (
+        # Install packages on which the stack is know to depend
+        # Note: it's not clear if agreement was reached to use all of these,
+        #       or they crept in by accident.
+
+        export PATH="$LSSTSW/miniconda/bin:$PATH"
```
jhoblitt: If PATH is re-declared in the subshell, does it need to be set for the rest of this script?

bin/deploy
```diff
@@ -7,45 +7,55 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
...
+        conda install --yes numpy scipy matplotlib requests cython sqlalchemy astropy
```
jhoblitt: What depends on cython and astropy? sqlalchemy is a technically a 3rd party package.

bin/deploy
```diff
@@ -7,45 +7,55 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
...
+        pip install stsci.distutils
```
jhoblitt: Is this for astropy?

bin/deploy
mjuric: Good point. I didn't notice it was declared on top as well.

I ran a test, and it looks like it's not needed (and looking at what the script is doing, there's nothing in here that should require anaconda).

I'll push a change to not set the PATH (which should solve #40 as well).

bin/deploy
mjuric: `palpy` and `pymssql`. sims packages (e.g., MAF) depend on `sqlalchemy` (they expect to have it via Anaconda).

See https://github.com/mjuric/conda-lsst/blob/master/bin/conda-lsst#L60-L66 for an (incomplete, as it doesn't include qserv) list.

bin/deploy
mjuric: `pyfits` (it otherwise silently pulls it in via `easy_install` when it's built).

mjuric: (n.b., we should get rid of `pyfits` -- woefully out of date).

mjuric:
Rebased to most recent master.

jhoblitt:
I'm attempting to create new jenkins build slaves using this branch.

jhoblitt:
A test jenkins instance has been building last night without incident.

PR #42, opened 2015-09-02
add repos-lint support for dict/hash/map format repo specification

jhoblitt:


etc/repos.yaml
```diff
@@ -1,3 +1,43 @@
...
+#
+#
+# format of this file
+# ---
+#
+# It is in [YAML](http://yaml.org/) format with a single top level
+# dict/hash/map.  Each keypair specifies a source repo and a specification of
+# how it may be cloned.  Key names must conform to the EUPS product name of the
+# source package.  Values may be either a URL (string) or a dict/hash/map with
+# *exactly* two keys: `url` and `branch`.  The second form generally should be
```
ktlim: It's "ref", not "branch".

bin/repos-lint
```diff
@@ -36,28 +53,42 @@ def validate_product_name(pkg):
...
+
+    if url.scheme == 'https':
+        urlopen(repo)
+    elif url.scheme == 'git':
+        # don't want to attempt to clone the repo as it might be large, so
+        # we're just checking for a valid hostname.  A better check might
+        # be to attempt a raw tcp connection.
+        socket.gethostbyname(url.netloc)
+    else:
+        raise ValueError('unsuported scheme: %s' % url.scheme)
```
timj: Typo: unsupported

etc/repos.yaml
jhoblitt: Nice catch -- I missed that when I renamed the key branch -> ref.

bin/repos-lint
jhoblitt: I've aspell'd the entire file and found a second exception string typo.

PR #49, opened 2015-10-18
(DM-4113) add an optional `lfs` key to repos.yaml hash format

jhoblitt:
The only allowed value is `true` to indicate the repo is using git-lfs.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -9,6 +9,7 @@ source ${SCRIPT_DIR}/../etc/settings.cfg.sh
 EUPS_VERSION=${EUPS_VERSION:-1.5.9}         # Version of EUPS to install
 MINICONDA_VERSION=${MINICONDA_VERSION:-latest} # Version of Miniconda to install
 GIT_VERSION=${GIT_VERSION:-2.2.2}           # Version of git to install
+LFS_VERSION=${LFS_VERSION:-1.0.2}           # Version of git-lfs to install
 
 set -e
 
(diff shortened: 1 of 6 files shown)


bin/repos-lint
```diff
@@ -37,42 +46,60 @@ def check_repo_spec(spec):
...
-    # currently we only support the url and ref keys
-    if all(key in spec for key in supported_keys):
+    if 'url' in spec:
         validate_repo_url(spec['url'])
+    if 'ref' in spec:
         validate_string(spec['ref'])
+    if 'lfs' in spec:
+        validate_bool(spec['lfs'])
+        if spec['lfs'] != True:
+            raise ValueError('the only supported value for lfs is true')
```
timj: Seems a bit mean for `false` not to be supported :smile:

jmatt:
I'm running into an build failure when I'm pip installing requirements using python3.5 on Mac OS X 10.10. This is relevant since ./bin/repos-lint uses python3

See: https://gist.github.com/jmatt/18f517d4037878fbfdb7

I had no issues when I installed PyYAML directly. I also had no issues with brew installed python3.4 (3.4.3_2). This came up when reviewing DM-4113. But is likely outside the scope of DM-4113, so maybe I should create another issue.

bin/deploy
```diff
@@ -74,6 +75,28 @@ test -f "$LSSTSW/lfs/.git.deployed" || ( # git
     touch "$LSSTSW/lfs/.git.deployed"
 )
 
+test -f "${LSSTSW}/lfs/.git-lfs.deployed" || (
```
jmatt: You are mixing your ${LSSTSW} and $LSSTSW formatted bash variables. Semantically they are the same unless there are issues with tokenizing. But it still sets off my OCD or mental linter. :)

jmatt:
+1

bin/repos-lint
```diff
@@ -37,42 +46,60 @@ def check_repo_spec(spec):
...
 
-    if missing_keys or unknown_keys:
+    if no_options or missing_keys or unknown_keys:
         message = ""
         for k in missing_keys:
             message += "missing mandatory key: %s\n" % k
         for k in unknown_keys:
             message += "unknown key: %s\n" % k
+        if no_options:
+            message += "aleast one of these keys is required: %s\n" \
```
ktlim: "at least"; you might also say that the other option is not to use the bare URL.

bin/repos-lint
jhoblitt: `false` works with `lsst_build` but the linter is intended to steer folks into using the minimal form.

bin/deploy
jhoblitt: I've added a new commit that improves the ${} consistency a bit.

bin/repos-lint
jhoblitt: fixed

jhoblitt:
This needs a new manual test before merging.

bin/deploy
jhoblitt: The cleanups were more noise then is reasonable in a single PR, and travis seems to be having  a problem installing shellcheck, so I've split it off as #53 which can be rebased after this PR is merged..

PR #51, opened 2015-10-22
Remove obs_cfht from repos.yaml

SimonKrughoff:
It turns out that obs_cfht has a dependency on a repo of test data.
I don't think we want to make lsst_apps depend on yet another data
repo, so I'm removing obs_cfht


ktlim:
I think it's OK for repos to be in `repos.yaml` but not in anything's table file.  Eventually we need to figure out a way to test these, even if they're as separate top-level products.

PR #60, opened 2015-11-30
no longer use install_name_tool to change load addresses

danielsf:
In the past, we have dealt with GalSim's inability to find anaconda's libpython2.7.dylib when building on a Mac by using install_name_tool to change the load address of libpython2.7.dylib after it is built by lsstsw's deploy script.  This change is not robust against users running conda update on their python installation (in which case, the load address can be reverted to its default value, which GalSim cannot find).  Recent changes to the EUPS GalSim package mean that we are now using install_name_tool to temporarily change the load addresses while GalSim is being built, and then reverting those changes after the build is complete.  The present pull request removes the use of install_name_tool from lsstsw's deploy script, as it is no longer needed.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -36,15 +36,6 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
     curl -# -L -O http://repo.continuum.io/miniconda/${miniconda_file_name}
     bash ${miniconda_file_name} -b -p "$LSSTSW/miniconda"
 
-    if [[ $(uname -s) = Darwin* ]]; then
-        #run install_name_tool on all of the libpythonX.X.dylib dynamic
-        #libraries in miniconda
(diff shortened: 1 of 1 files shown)


timj:
@danielsf given the GalSim discussion, are you saying that modifying this patch to add the `@rpath` would fix the GalSim problem?

danielsf:
Yes (in the case of lsstsw).  But, based on the discussion on SIM-1637, we need to find a solution that does not rely on install_name_tool anyway, right?

timj:
Yes. We can't insist on an anaconda installed by us so of course none of these hacks can be the solution.

jhoblitt:
@timj / @danielsf Should we go ahead and merge this?

timj:
Yes. We can't rely on hacking up the installed python libraries.

jhoblitt:
Works for me.

PR #63, opened 2015-12-03
(DM-4534) shellcheck

jhoblitt:


jhoblitt:
Currently based on top of #62 -- should be rebased on master after that PR is merged.

bin/deploy
```diff
@@ -33,15 +33,15 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
...
-    curl -# -L -O http://repo.continuum.io/miniconda/${miniconda_file_name}
-    bash ${miniconda_file_name} -b -p "$LSSTSW/miniconda"
+    curl -# -L -O "http://repo.continuum.io/miniconda/${miniconda_file_name}"
+    bash "$miniconda_file_name" -b -p "${LSSTSW}/miniconda"
 
     if [[ $(uname -s) = Darwin* ]]; then
         #run install_name_tool on all of the libpythonX.X.dylib dynamic
         #libraries in miniconda
-        for entry in $LSSTSW/miniconda/lib/libpython*.dylib
+        for entry in ${LSSTSW}/miniconda/lib/libpython*.dylib
```
timj: So there is no need to use `"${LSSTSW}"/` here?

jhoblitt: Good question... perhaps a parser bug?

jhoblitt: Yep. Known bug: https://github.com/koalaman/shellcheck/issues/300

bin/deploy
```diff
@@ -171,7 +174,7 @@ cat <<EOF
 
 Done. Run the following:
 
-    . $LSSTSW/bin/setup.sh
+    . ${LSSTSW}/bin/setup.sh
```
timj: Add quotes around `${LSSTSW}`?

timj:
Looks ok to me.

bin/deploy
jhoblitt: The variable interpolation is inside a heredoc so it is "quoted".

timj: I know. It's what you want people to copy and paste into their shell. So technically they would need the quotes wouldn't they?

jhoblitt: Good point.

bin/deploy
```diff
@@ -33,15 +33,15 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
 
     miniconda_file_name="Miniconda-${MINICONDA_VERSION}-${ana_platform}.sh"
     echo "::: Deploying Miniconda ${MINICONDA_VERSION} for ${ana_platform}"
-    curl -# -L -O http://repo.continuum.io/miniconda/${miniconda_file_name}
-    bash ${miniconda_file_name} -b -p "$LSSTSW/miniconda"
+    curl -# -L -O "http://repo.continuum.io/miniconda/${miniconda_file_name}"
+    bash "$miniconda_file_name" -b -p "${LSSTSW}/miniconda"
 
     if [[ $(uname -s) = Darwin* ]]; then
```
mjuric: I thought we got rid of this -- I think GalSim can now build w/o this hack (at least on pre-El Capitan versions).

The problem is that this may make a build succeed on `lsstsw`, where it would fail with vanilla Anaconda. Maybe open a new PR?

mjuric:
Seems OK, though I'm not sure about too rigorous `{}`-ing of variable names (e.g.,` ${LSSTSW}` -- someone less experienced may think there's some special magic happening there).

I also always die a bit inside when I see space-indented bash code :(. If God (ok, Chet Ramey... or maybe Brian Fox?) intended spaces to be used, he'd make `<<-FOO` heredocs ignore leading spaces, not just tabs :).

That said, OK to merge.

bin/deploy
timj: That's in PR #60 which is still open.

jhoblitt:
I've rebased on current master but I'm not intending to merge this week (I remain concerned about breakage).  @mjuric I generally try to use ${} inside of all double quotes for consistency (the linter for puppet requires this, so I've been training into the habit) but I'm happy to follow a common style.

jhoblitt:
I should be more specific, I use ${} inside of double quotes if there is anything other than just the variable inside the quoted string.

jhoblitt:
and it looks like my kludge to work around travis not having the shellcheck package in their repos has broken anyways...

timj:
Is this a dead PR?

jhoblitt:
Superseded by #212

PR #82, opened 2016-02-10
pin conda package versions

jhoblitt:


bin/deploy
```diff
@@ -47,15 +81,20 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
 
     (
         # Install packages on which the stack is know to depend
```
timj: "known" not "know"

bin/deploy
```diff
@@ -47,15 +81,20 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
...
         # XXX the list of conda packages should be kept in sync with
         # https://github.com/lsst/miniconda/blob/master/ups/eupspkg.cfg.sh
         # The lsstsw list of packages should be considered authoritative
-        conda install --yes numpy scipy matplotlib requests cython sqlalchemy astropy pandas
-        pip install stsci.distutils
+        export PATH="$LSSTSW/miniconda/bin:$PATH"
+
+        if [[ $BLEED_DEPLOY == true ]]; then
+            conda install --yes nomkl numpy scipy matplotlib requests cython sqlalchemy astropy pandas
+            pip install stsci.distutils
```
timj: This pip install is not needed. There is an eups package for it.

bin/deploy
```diff
@@ -47,15 +81,20 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
...
+        else
+            conda install --yes --file "${SCRIPT_DIR}/../etc/conda_packages.txt"
+            # stsci.distutils setup_requires d2to1
+            pip install stsci.distutils==0.3.7 d2to1==0.2.12.post1
```
timj: Now eups package. Remove.

bin/deploy
```diff
@@ -47,15 +81,20 @@ test -f "$LSSTSW/miniconda/.deployed" || ( # Anaconda
...
-        export PATH="$LSSTSW/miniconda/bin:$PATH"
         # XXX the list of conda packages should be kept in sync with
         # https://github.com/lsst/miniconda/blob/master/ups/eupspkg.cfg.sh
         # The lsstsw list of packages should be considered authoritative
-        conda install --yes numpy scipy matplotlib requests cython sqlalchemy astropy pandas
-        pip install stsci.distutils
+        export PATH="$LSSTSW/miniconda/bin:$PATH"
+
+        if [[ $BLEED_DEPLOY == true ]]; then
+            conda install --yes nomkl numpy scipy matplotlib requests cython sqlalchemy astropy pandas
```
timj: Should be a comment somewhere indicating that we are deliberately building without MKL, even if it's just in the commit message.

bin/deploy
jhoblitt: Outside of my changeset but I'll fix it. :)

bin/deploy
jhoblitt: I will remove it since the dep is tracked in the table files now.  However, I will note that I dislike the behavior of the "vendored" EUPS python packages as setuptools downloads deps from pypi and exactly what is retrieved could change based on which packages are setup n the env.  Also, the dep versions are free floating.

timj: in this particular case `stsci.distutils` will not download anything else so the EUPS package is already handling everything.

jhoblitt:
I've updated one of the comment blurbs to note that the miniconda2 package pulls the conda package list from this repo.

PR #85, opened 2016-02-20
Mark ci_hsc as git-lfs.

wmwv:
Originally just listed ci_hsc as a simple repo.  Forgot that I needed to list it as a git-lfs enabled repo.  This commit adds that.

### DIFF
--- a/etc/repos.yaml
+++ b/etc/repos.yaml
@@ -224,4 +224,6 @@ validation_data_decam:
   lfs: true
 pyyaml: https://github.com/lsst/pyyaml.git
 requests: https://github.com/lsst/requests.git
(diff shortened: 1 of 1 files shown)


jmatt:
@wmwv Just a note that git-lfs + lsst-build may not work with already existing repositories. See: https://jira.lsstcorp.org/browse/DM-4733

wmwv:
Yes, this turned out to work fine.  I had just added it to the etc/repos.yaml a few minutes before and then realized I hadn't marked it as git-lfs.

https://ci.lsst.codes/job/stack-os-matrix/label=centos-6/lastBuild/console

PR #117, opened 2016-08-04
update exit status to indicate if prepare completed

jhoblitt:


bin/rebuild
```diff
@@ -16,6 +16,25 @@
...
+#
+# Add 128 to non-zero exit statuses if the `lsst-build prepare` subcommand has
+# completed.  This is to allow consumer scripts to be able to determine if a
+# failure in this script occurred before or after the prepare step.
+#
+cleanup() {
+    local rv=$?
+
+    [[ $rv == 0 ]] && exit
+    [[ $PREPARED == true ]] && rv=$((rv + 128))
```
ktlim: Exit status is mod 256, I believe.  Are you sure `$?` won't ever be >128 to start with?

jhoblitt: I thought about sanity checking it but didn't bother since 2 is the highest value in usage.  I'll fix it up.

PR #118, opened 2016-08-05
split miniconda & conda pkg install into separate steps

jhoblitt:
Prior to this change, if `deploy` successfully installed miniconda but a
failure occurred while installing conda packages, the install could not
be repaired by rerunning `deploy`. This was due to the miniconda
installer refusing to run if the install directory already exists.  This
is avoided by protecting the miniconda and conda package installs with
separate lockfiles.


jhoblitt:
Tested as working in a jenkins sandbox env when started from a clean workspace.  It can not recover from an existing installation failure as the lockfile for the miniconda install will not be present.

PR #121, opened 2016-08-24
    disable conda install progress bar when there is no tty

jhoblitt:
The package installation progress bar is causing 'scroll shock' in the
travis/jenkins output logs.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -102,13 +102,25 @@ test -f "$LSSTSW/miniconda/.packages.deployed" || ( # conda packages
     # uses the conda package specification from this repo.
     export PATH="$LSSTSW/miniconda/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


timj:
I'm surprised that conda can't work out if it's attached to a tty and disable it itself.

jhoblitt:
Actually, we should just test for a tty rather than turning it off unconditionally.  I'll fix it up.

timj:
@jhoblitt @parejkoj is reporting that the `tty` command is failing on OS X. `--quiet` is not supported and needs to be `-s`.

jhoblitt:
Aye.  I am tempted to take the OSX tests out of the travis allowed failures list.  My hesitation is that the OSX build infrastructure keeps breaking.

PR #122, opened 2016-08-25
Stop masking healpy and scipy packages

timj:
The healpy and scipy packages build fine. No reason to mask them any more.


jhoblitt:
This looks good.  I'm unsure if there is a `condor` package referenced from any eups product.  I know that `cuda_toolkit` is referenced from afw.

PR #126, opened 2016-09-08
setup.csh now works like other setup files

kfindeisen:
`setup.csh` can now be sourced from login shell scripts and will set up lsstsw correctly. Support for directly sourcing the script has been kept for backward-compatibility with existing development environments.

@djreiss, can you confirm that your preferred way of executing the script still works?

Note to admins: I know forking is against LSST policy, but my account hasn't been added to the LSST organization yet. Apologies for the (hopefully) one-time violation!

### DIFF
--- a/bin/setup.csh
+++ b/bin/setup.csh
@@ -1,9 +1,27 @@
 # setup lsstsw environment
 #
-# source this file from your location of <lsstsw>
+# source this file from your ~/.cshrc
 #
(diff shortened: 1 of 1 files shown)


kfindeisen:
Found a bug in the new script (it always creates a file called `2` in the lsstsw directory), I'll push directly to the branch once it's fixed.

PR #127, opened 2016-09-08
DM-7552: Add support for more shells

timj:


bin/setup.csh
```diff
@@ -0,0 +1,46 @@
...
+# source this file from your ~/.cshrc
+#
+# possibly blank, but can't be parsed as $_
+set cmd=($_)
+if ( $?0 ) then
+    set source=$0
+else if ( $#cmd >= 3 ) then
+    set source=`pwd`/${cmd[2]}
+else if (-f /usr/sbin/lsof ) then
+    set source=`/usr/sbin/lsof +p $$ | grep -oE /.\*setup.csh`
```
timj: why do we need `lsof`? `newinstall` doesn't need it.

bin/setup.csh
```diff
@@ -0,0 +1,46 @@
...
+endif
+
+setenv PATH "$LSSTSW/miniconda/bin:$PATH"
+setenv PATH "$LSSTSW/lfs/bin:$PATH"
+setenv PATH "$LSSTSW/bin:$PATH"
+rehash
+
+setenv MANPATH "$LSSTSW/lfs/share/man:"
+
+source $LSSTSW/eups/current/bin/setups.csh
```
timj: Please add double quotes. `$LSSTSW` may have spaces in it.

bin/setup.csh
```diff
@@ -0,0 +1,46 @@
...
+
+setup -r $LSSTSW/lsst_build
```
timj: Add quotes.

bin/setup.zsh
```diff
@@ -0,0 +1,24 @@
...
+    return
+fi
+
+export PATH="$LSSTSW/miniconda/bin:$PATH"
+export PATH="$LSSTSW/lfs/bin:$PATH"
+export PATH="$LSSTSW/bin:$PATH"
+
+export MANPATH="$LSSTSW/lfs/share/man:"
+
+. $LSSTSW/eups/current/bin/setups.sh
```
timj: Add quotes.

bin/setup.zsh
```diff
@@ -0,0 +1,24 @@
+# setup lsstsw environment
+#
+# source this file from your ~/.bashrc
+#
+# relative to <lsstsw>/bin/
+LSSTSW=$(cd "$(dirname "${(%):-%x}")/.."; pwd)
```
timj: newinstall uses:
```zsh
LSST_HOME=\`dirname "\$0:A"\`
```
which looks like it is simpler.

bin/setup.zsh
```diff
@@ -0,0 +1,24 @@
...
+
+if [[ ! -f $LSSTSW/eups/current/bin/setups.sh ]]; then
```
timj: Quotes. `$LSSTSW` might include spaces.

bin/setup.zsh
```diff
@@ -0,0 +1,24 @@
...
+
+export PATH="$LSSTSW/miniconda/bin:$PATH"
+export PATH="$LSSTSW/lfs/bin:$PATH"
+export PATH="$LSSTSW/bin:$PATH"
+
+export MANPATH="$LSSTSW/lfs/share/man:"
+
+. $LSSTSW/eups/current/bin/setups.sh
+
+setup -r $LSSTSW/lsst_build
```
timj: Quoting.

timj:
Are there no ksh fans out there?

bin/setup.csh
kfindeisen: `newinstall` is run from the LSST root directory (once?). Both `setup.sh` and the new scripts are designed to set up environment variables automatically when the user logs in, so it can't rely on the user executing from a specific location on the build path.

So that entire ugly tree is the best csh emulation I can find for `$BASH_SOURCE`

timj:
The only difference between the zsh and bash version (hmm, should `setup.sh` be renamed `setup.bash`?) is the setting of LSSTSW line. Can bash/zsh scripts share work out which of those lines to use. Would be good if it was a single script that tried different options.

bin/setup.csh
timj: You may find in the future that you really don't want to have the setup script sourced as part of your log in. I define a couple of shell aliases to let me initialize different `lsstsw` build trees (mainly for python versions) in different shells (switching trees later on is hard as you have to work out which environment variables to clean up).

Remember that `newinstall` creates the `loadLSST.*` setup scripts which are run to initialize the LSST environment. The fact that the code is in `newinstall` which is called once is not relevant to the discussion.

This is the logic from `loadLSST.csh` that is equivalent to what is being attempted here:
```csh
               set sourced=(\$_)
                if ("\${sourced}" != "") then
                   # If not already initialized, set LSST_HOME to the directory where this script is located
                   set this_script = \${sourced[2]}
                   if ( ! \${?LSST_HOME} ) then
                          set LSST_HOME = \`dirname \${this_script}\`
                          set LSST_HOME = \`cd \${LSST_HOME} && pwd\`
                   endif
```

kfindeisen: Maybe so, but `$LSSTSW/bin/setup.sh` is already designed as a login script. If you want to change whether it's a login or call-only script, we'd also have to modify the bash version of the script so that it agrees.

As for your last example, `$_` only contains useful information if you've directly sourced the script from the command line. It's blank if the script was sourced from another script. :frowning: Though I agree that code handles the relative- vs. absolute-path question better and I'll change the parsing accordingly.

jonathansick:
I'm trying this out now—this will make my life 💯 better, thanks.

One detail is that the `./bin/deploy` script still only says

```
Done. Run the following:

    . {{path}}/lsstsw/bin/setup.sh

to begin using it.
```

Either list all supported setup.sh-equivalents here, or provide context-specific advice to the user.

bin/setup.csh
timj: I really hadn't expected people to be adding these to their `.login` file but if you have to do something more complicated to support that working then please ignore my comments. Maybe add a comment to the file though to explain the use case.

timj:
Changing the deploy output would be good.

kfindeisen:
Fixed agreed-upon changes, and rebased all my commits in the process.

timj:
Is @ctslater going to fix the `.zsh` version?

timj:
The day you start using both a python 3 lsstsw and a python 2 lsstsw is the day you remove it from your `.cshrc` file. :smile:

timj:
@ctslater Following up on my previous `zsh` comment. I'd prefer it if `setup.sh` worked for both zsh and bash given how close they are. It seems that you can check for the `$ZSH_NAME` or `$BASH` environment variable and just tweak the logic accordingly.

See http://stackoverflow.com/questions/3327013/how-to-determine-the-current-shell-im-working-on

kfindeisen:
Hmm... the `deploy` script currently uses `$SHELL` to figure out what it's running on. This works on tcsh -- the script recommends `setup.csh` despite being run as bash -- but that discussion says it may not always work...

bin/deploy
```diff
@@ -236,12 +236,19 @@ test -f "$LSSTSW/lsst_build/.deployed" || ( # Clone lsst_build
...
-
-Done. Run the following:
-
-    . $LSSTSW/bin/setup.sh
-
-to begin using it.
+echo "Done. Run the following:"
+echo 
+
+if [[ $SHELL =~ bash$ ]]; then
```
timj: Does this work? Won't `$SHELL` always be `bash` because this script is always run with `bash`? I don't think it's possible for this script to know what your login shell is. I think you just list all the options and let people choose. The options are enough to make people aware that they have a choice.

kfindeisen: I tested it on tcsh and it works.

timj: Oh right, `$SHELL` is not your current shell it's your login shell. Sorry. I had confused myself into thinking it was the current shell (but it clearly isn't given the other discussion regarding `$ZSH_NAME`).

bin/setup.sh
```diff
@@ -1,14 +1,22 @@
...
+# source this file from your ~/.bashrc or ~/.zshrc
 #
 # relative to <lsstsw>/bin/
-LSSTSW=$(cd "$(dirname "$BASH_SOURCE")/.."; pwd)
-
-if [[ ! -f "$LSSTSW/eups/current/bin/setups.sh" ]]; then
-    echo "error: eups not found in $LSSTSW/eups/current" 1>&2
-    echo "  you may need to [re]run bin/deploy to [re]deploy EUPS." 1>&2
-    return
+if [[ -z $ZSH_NAME ]]; then
```
timj: in this if block just define $LSSTSW and the file extension of the EUPS `setups.[z]sh` file. Then just use  `$SUFFIX` in later code. That will minimize all the duplication. Ironically, EUPS generates identical sh and zsh setup files.

jmeyers314: Fair enough.  Done.

bin/setup.csh
```diff
@@ -0,0 +1,46 @@
...
+
+if ( $?source ) then
+    set LSSTSW=`dirname $source`
+    set LSSTSW=`cd $LSSTSW/.. && pwd`
+endif
+
+if ( ! $?LSSTSW ) then
+    echo "error: could not figure out LSSTSW directory"
+    echo '  you can specify the directory by setting $LSSTW in your ~/.cshrc'
+    exit 1
```
timj: is `exit` okay in tcsh when you are sourcing a shell script? On bash it would exit your entire shell.

ctslater:
Thanks for taking care of the rest of the zsh script @jmeyers314.

bin/setup.csh
kfindeisen: Yeah, that makes me nervous too. But it seems to work (both sourced and executed), and I don't know of any other equivalents to bash `return`.

PR #136, opened 2016-09-29
DM-7676: lsof used by setup.csh not always in /usr/sbin

kfindeisen:
`setup.csh` has been modified to look for `lsof` on the path, falling back to hard-coded locations only if necessary. [The issue](https://jira.lsstcorp.org/browse/DM-7676) was reported on Ubuntu 14.04, but I have only tested the fix on Ubuntu 16.04.

### DIFF
--- a/bin/setup.csh
+++ b/bin/setup.csh
@@ -2,16 +2,20 @@
 #
 # source this file from your ~/.cshrc
 
-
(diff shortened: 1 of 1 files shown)


ktlim:
Looks OK.  Might be nice to extract the `lsof` command line so that it doesn't have to be maintained in two places.

PR #143, opened 2016-11-09
Adding sims_survey_fields to repos.yaml

mareuter:
I've tested this locally and it seems to work. BTW, the Travis builds seem to be broken on SSL connection to the Linux kernel site.


jhoblitt:
The ssl failure is due to a problem with the travis container env.  I opened a lsst jira and a github issue on travis several weeks ago.

PR #144, opened 2016-11-18
switch travis to trusty beta containers

jhoblitt:
To work around curl failing to download git sources from kernel.org due
to not properly supporting the TLS cert.


jhoblitt:
Closing as this was a failed test.

PR #152, opened 2016-12-19
update conda env versions

jhoblitt:


bin/deploy
```diff
@@ -112,10 +110,23 @@ test -f "$LSSTSW/miniconda/.packages.deployed" || ( # conda packages
...
-        # The conda Intel MKL linked packages are intentionally avoided.
-        # See: https://jira.lsstcorp.org/browse/DM-5105
-        CONDA_BLEED_PACKAGES=(nomkl numpy scipy matplotlib requests cython sqlalchemy astropy pandas future)
+        # lsst_distrib / lsst_sims deps
+        CONDA_BLEED_PACKAGES=(numpy scipy matplotlib requests cython sqlalchemy astropy pandas future)
         ARGS+=("${CONDA_BLEED_PACKAGES[@]}")
+
+        # matplotlib pulls in qt and there are concerns about qt5 causing
+        # breakage, so we are pinning to 4.x
+        ARGS+=("pyqt==4.11.4")
```
timj: I disagree with this. I think we should definitely pin `pyqt` at v4 when we request a version pinned build but if I'm doing a bleeding edge build I want to know what breaks with the bleeding edge. I don't want to get confused by pinning regardless.

bin/deploy
```diff
@@ -112,10 +110,23 @@ test -f "$LSSTSW/miniconda/.packages.deployed" || ( # conda packages
...
+
+        # The conda Intel MKL linked packages are intentionally avoided [on
+        # linux].
+        # See: https://jira.lsstcorp.org/browse/DM-5105
+        if [[ $(uname -s) == Linux* ]]; then
+            ARGS+=("nomkl")
+        fi
+
+        # lsst_build requires pyyaml
+        ARGS+=("pyyaml")
```
timj: If you are being pedantic about listing `lsst_build` dependencies I think it needs the `future` package and in theory you can remove `future` from the list of products needed by `lsst_apps/distrib` above because we ship our own EUPS future.

bin/deploy
timj: I want bleed to be bleed. When you are coming up with a new package list then forcing it in there temporarily is fine with me. I just don't want bleed to not be bleed just because we are waiting on Continuum for this particular problem to be fixed.

jhoblitt:
@timj Comments have been addressed.

PR #156, opened 2017-01-19
DM-6058: Add numexpr and bottleneck to package list

danielsf:
Add numexpr and bottlenecks to the list of conda packages automatically installed by lsstsw.

Also: add the pandas EUPS stub to repos.yaml

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -112,7 +112,7 @@ test -f "$LSSTSW/miniconda/.packages.deployed" || ( # conda packages
     if [[ $BLEED_DEPLOY == true ]]; then
         PACKAGES=()
         # lsst_distrib / lsst_sims deps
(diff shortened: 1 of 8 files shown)


jhoblitt:
@timj I am OK with this PR but rather it isn't merged until my question on https://github.com/lsst/pandas/pull/1 is answered.

timj:
@danielsf I think you are missing the patch to the deploy script to add `numexpr` etc to the list for the bleeding edge install.

danielsf:
@timj is that not controlled by `conda_packages-[linux, osx]-64.txt`?  Which piece of code do I need to patch (I'm only vaguely aware of how lsstsw works)

timj:
Those text files are for pinning the versions. When we want to float up to the most recent versions for testing we bypass those and specify floating versions of all packages. See the `PACKAGES` variable in `bin/deploy` (search for `pandas`).

PR #159, opened 2017-01-24
support for installing eups from git

jhoblitt:


jhoblitt:
@timj The plan is to leave it alone until it breaks -- it should be a pretty obvious failure when only that matrix slot blows up.

PR #160, opened 2017-01-25
update "lsst-dev"'s FQDN

jhoblitt:
`lsst-dev` is now an alias to `lsst-dev01`.


jhoblitt:
travis seems to be bogged down again today so I'm going to "blind" merge this as it is purely a change to "data".

PR #163, opened 2017-01-30
refactor versiondb "push" support

jhoblitt:


jhoblitt:
Travis, for unknown reasons, seems to no longer be building this repo and I'm unable to access the org configuration at all (times out).  This PR passed as a branch in a personal fork.

https://travis-ci.org/jhoblitt/lsstsw/builds/197061679

I am squashing some commits now but not changing the code.

.travis.yml
```diff
@@ -74,6 +98,69 @@ script: |
...
+
+        mkdir -p "$tmpdir"
+        git clone "$VERSIONDB_REPO" "$tmpdir"
+        cd $tmpdir
+        # it appears that with some versions of git, --author doesn't bypass
+        # the user.email/user.name check.
+        git config --local user.email 'author@example.com'
+        git config --local user.name 'A U Thor'
+        git commit --allow-empty -m 'initial commit' --author='A U Thor <author@example.com>'
+        git push origin master
```
athornton: I presume we have an ssh key in the user's environment or something?

.travis.yml
```diff
@@ -74,6 +98,69 @@ script: |
...
+      rebuild cfitsio
+
+      cd $VERSIONDB_REPO
+      if [[ $VERSIONDB_PUSH == true ]]; then
+        echo 'should have master ref'
+        git show-ref --heads | grep -q refs/heads/master
+
+        revs=($(git rev-list master))
+
+        # the first commit is a garage commit in order to create the master ref
```
athornton: Do you mean "garbage" here?

.travis.yml
```diff
@@ -31,6 +39,22 @@ matrix:
   allow_failures:
     - os: osx
 script: |
+  set -e
```
athornton: The set -e, I presume, is so that you explode in the middle of the git statements at the first failure rather than trying to run subsequent ones?

jhoblitt: Essentially, yes.  I think travis executes the script with `-e` set by default but I like to be explicit.

.travis.yml
jhoblitt: This test is using the local filesystem so ssh keys aren't needed.

.travis.yml
jhoblitt: The `git commit --allow-empty -m 'initial commit'` commit so that `git pull` doesn't blow up because there aren't any refs in the remote to fetch.  Do you have a suggestion on how to improve the comment?

.travis.yml
athornton: So origin is local?  I missed that.

.travis.yml
```diff
@@ -123,7 +124,7 @@ script: |
 
       ./bin/deploy
 
-      (cd versiondb; mkdir dep_db ver_db manifests)
+      mkdir -p versiondb/{dep_db,ver_db,manifests}
```
athornton: Fancy!

.travis.yml
jhoblitt: The origin is ` VERSIONDB_REPO=./versiondb-test`

.travis.yml
jhoblitt: Removed a subshell.

PR #164, opened 2017-02-14
add sky_brightness_pre repo

yoachim:
yup.

### DIFF
--- a/etc/repos.yaml
+++ b/etc/repos.yaml
@@ -96,6 +96,7 @@ gsl: https://github.com/lsst/gsl.git
 sims_maf: https://github.com/lsst/sims_maf.git
 sims_skybrightness: https://github.com/lsst/sims_skybrightness.git
 sims_skybrightness_data: https://github.com/lsst/sims_skybrightness_data.git
+sims_skybrightness_pre: https://github.com/lsst/sims_skybrightness_pre.git
 lsst: https://github.com/lsst/lsst.git
 boost: https://github.com/lsst/boost.git
 sims_sed_library:


danielsf:
Does sims_skybrightness_pre actually need to use syseng_throughputs as opposed to just throughputs?  syseng_throughputs is not currently distributed with lsstsw, so running `lsstsw/bin/rebuild` fails because it does not know from where to clone syseng_throughputs.  I'm not sure if there is an intentional reason we are not distributing syseng_throughputs (@rhiannonlynne?)

Also: the utils.py script in sims_skybrightness_pre that actually uses syseng_throughputs imports a `bandpassUtils` module.  I do not see that anywhere.  Is it something you forgot to push up to sims_skybrightness_pre?

rhiannonlynne:
bandpassUtils is something in syseng_throughputs (python dir) that basically lets you get to the 'consumer-level' throughputs curves that are available in throughputs.

danielsf:
Okay.  Then maybe we should make sims_skybrightness_pre use throughputs directly, instead...?

rhiannonlynne:
Why is there python code in both the sims_skybrightness_pre/python  and in sims_skybrightness_pre/python/lsst/sims/skybrightness_pre directories? Is there some logical separation between those pieces that I'm not understanding? and if so, should the stuff directly in python live in the examples dir? (should these comments actually be reviews on sims_skybrightness_pre?)

rhiannonlynne:
Actually .. so I think maybe I know why sims_skybrightness_pre is using syseng_throughputs + bandpassUtils, which is to calculate a full m5 scaling value? (Cm / dCm_infinity / kAtm). This is available in the python code which goes with syseng_throughputs, but is not currently housed elsewhere.
Perhaps this is something we should add to/move to sims_photUtils ??

danielsf:
I am not opposed to making syseng_throughputs available to the world.  I just know that we currently do not.  Are you opposed to making it available?

rhiannonlynne:
I thought the deal was that distributable code was supposed to go in the "LSST" organization. Non distributable code was elsewhere. The syseng_throughputs repo is pretty firmly in the lsst-pst organization, unless you talk to Chuck and Wil about it (for various control/access reasons). This is my only objection .. I guess I think if we make it distributable we should consider removing the 'throughputs' repo though?

danielsf:
Which do we think would cause more general consternation: distributing syseng_throughputs and getting rid of throughputs, or moving the m5 calculation to sims_photUtils (as far as I know, sims_photUtils is still stripped down enough that you can just clone and run it if you don't want to go through the full `eups distrib install`/`lsstsw` rigamarole)?

yoachim:
ok, I'm back from vacation! You guys went to town while I was gone. 

Looks like I just forgot to remove the utils.py, that's just another copy of the m5 code that's now in sims_utils. I'll clean that up and remove the syseng dependency.

PR #168, opened 2017-02-28
bump eups to 2.1.1

jhoblitt:


jhoblitt:
Travis is a broken state due to the ongoing s3 problems.  I tested this in a container by running `deploy` with the EUPS version specified as `2.0.2` and then reruning it with `2.1.1`.  `deploy` was able to upgrade the eups installation.

PR #173, opened 2017-03-14
bump EUPS version to 2.1.2

jhoblitt:


timj:
@jhoblitt this PR seems to be much more than would be implied by the PR title and the lack of comment. It seems to be also including an update of #63 -- did you mean to do that?

jhoblitt:
The shellcheck/style fixes have been split out into #174

PR #180, opened 2017-04-07
cleanup conda lock files before bootstraping conda env

jhoblitt:
This failure mode has been encountered when re-running deploy after an
unsuccessful installation attempt.

    LockError: Lock error: Already locked: Lock error:
    LOCKERROR: It looks like conda is already doing something.
    The lock [u'.../lsstsw/miniconda/pkgs/mkl-11.3.3-0.tar.bz2.pid86953.conda_lock'] was found. Wait for it to finish before continuing.
    If you are sure that conda is not running, remove it and try again.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -129,6 +129,11 @@ test -f "${LSSTSW}/miniconda/.packages.deployed" || ( # conda packages
     # shellcheck disable=SC2031
     export PATH="$LSSTSW/miniconda/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


athornton:
Do you want some check to make sure that conda isn't actually running?  Or is the context such that it really, really never should be?

jhoblitt:
The PID is encoded in the lockfile name but conda is obviously not checking for it (the PID recorded has never been running when I have investigated). https://github.com/conda/conda/issues/4862

PR #186, opened 2017-06-15
bump eups version to 2.1.3

jhoblitt:


jhoblitt:
A manual test was able to build py2/centos-6, py2/centos-7, and py3/centos-6.

PR #190, opened 2017-07-10
misc. minor cleanups

jhoblitt:
These are few minior cleanups desired before refactoring the repos.yaml
handling.

### DIFF
--- a/bin/mass-tag
+++ b/bin/mass-tag
@@ -8,7 +8,7 @@ if [[ -z $1 ]]; then
     echo
     echo "     -d : instead of setting it, delete the tag"
     echo
-    echo "WARNING: THIS WILL APPLY $tag TO ALL LSST/DMS REPOSITORIES IN ./build"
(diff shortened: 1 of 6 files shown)


bin/mass-tag
```diff
@@ -29,19 +29,19 @@ cd build
...
 for gg in */.git; do
     (
-        r=$(dirname $gg)
-        cd $r
+        r=$(dirname "$gg")
+        cd "$r"
 
         ORIGIN=$(git config --get-all remote.origin.url | grep "LSST/DMS/" | sed 's|git://git.lsstcorp.org/|git@git.lsstcorp.org:|')
-        [[ -z $ORIGIN ]] && continue
+        [[ -z $ORIGIN ]] && exit # continue looop
```
timj: "looop". `exit` continues the loop?

jhoblitt: Yes, it returns from the subshell which is nested in the loop body.

PR #197, opened 2017-08-25
DM-11646: Change default to python 3

timj:


timj:
@jhoblitt I don't entirely understand why it's failing to build cfitsio on linux but is fine on mac.

jhoblitt:
The travis failure is a recent change in probably related to the update of the debian version in the docker image. At least the git build is no longer failing.  There is an open ticket to change these tests to run in our own docker image.

PR #200, opened 2017-09-06
bump git version to 2.14.1

jhoblitt:
A newer version of git is required to support $GIT_SSH_COMMAND.  This
env var would be a simple way to disable StrictHostKeyChecking when
cloning lsst/versiondb via git (only used for the "canonical" build env
under jenkins).

See:

https://stackoverflow.com/questions/7772190/passing-ssh-options-to-git-clone


jhoblitt:
Travis stll hasn't run (after a day)... merging anyways.

PR #202, opened 2017-09-06
DM-11727: Update pinned packages

timj:
* Add cmake to build deps
* Updates miniconda base version
* Updates all pinned packages to current versions (this includes a setuptools update).

I am not trying to be clever with cmake being installed only when jenkins is detected.


timj:
I've removed the cmake component of this PR.

timj:
I've removed `conda` and `conda-env` from the pinned list.

PR #203, opened 2017-09-14
Build python package in test and not rely on C compiler

timj:


timj:
Why is the gcc version so old on travis trusty? Maybe we should switch the rebuild test to something trivial like `rebuild python` since the point is to test that `rebuild` works with the install, not that the compiler is new.

jhoblitt:
Note that trusty was tried in #144 and didn't work.  The trusty image has likely been updated since then, so probably worth trying.

timj:
trust is still using ancient gcc 4.8.4.

jhoblitt:
I think the build probably be changed to use `docker.io/lsstsqre/centos:7-stackbase`

timj:
For some reason the travis job failed because it was trying the previous commit not the current HEAD.

timj:
@jhoblitt what do you think? This shows it was an issue with the compiling and still tests that `rebuild` is working properly.

timj:
@jhoblitt I'll remove the trusty change before merging.

jhoblitt:
This seems reasonable but I still feel the ultimate fix is to test inside our own docker image.

PR #204, opened 2017-09-15
completely remove osx from travis matrix

jhoblitt:


timj:
I'm trying to investigate the problem with the failures on #203.

jhoblitt:
Lets deal with the build problems separately -- which has actually been multiple independent problems with various iterations of the travis build env.

timj:
Yes. I'm working on it. I'm going to make it build something that doesn't rely on very old compilers.

PR #206, opened 2017-10-19
DM-12266: Fix disabling of MKL

pschella:


bin/deploy
```diff
@@ -255,6 +248,16 @@ main() {
 
     conda "${ARGS[@]}"
 
+    # The conda Intel MKL linked packages are intentionally avoided [on linux].
+    # See: https://jira.lsstcorp.org/browse/DM-5105 and
+    #      https://jira.lsstcorp.org/browse/DM-12266
+    if [[ $BLEED_DEPLOY == true ]]; then
+      if [[ $(uname -s) == Linux* ]]; then
+        conda install --yes nomkl
```
jhoblitt: If the intent is to install the `nomkl` package before anything else is installed, the ordering is wrong -- this probably should be at ~ line 235.

bin/deploy
```diff
@@ -255,6 +248,16 @@ main() {
...
+        conda remove --yes mkl mkl-service
```
jhoblitt: Is this needed if only `nomkl` packages are installed?

jhoblitt:
I believe this PR is no longer relevant.

PR #208, opened 2017-11-07
remap validation data products with large lfs repos

jhoblitt:


jhoblitt:
Indeed - however, It was only a few lines so I decided it was worth the larger diff to keep the formatting consistent.

PR #214, opened 2018-04-13
add -d (debug) flag to publish script

jhoblitt:


jhoblitt:
This is needed to debug a production problem with publishing, thus it is being merged without review.

PR #218, opened 2018-07-02
add conda "bleed" mode environment files

jhoblitt:
Separates the conda package "data" from the deploy "code" so that the canonical "bleed" package list may be modified without touching "code".

based on #220

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -10,7 +10,9 @@ source "${SCRIPT_DIR}/../etc/settings.cfg.sh"
 EUPS_VERSION=${EUPS_VERSION:-2.1.4}         # Version of EUPS to install
 EUPS_GITREV=${EUPS_GITREV:-""}
 EUPS_GITREPO=${EUPS_GITREPO:-https://github.com/RobertLuptonTheGood/eups.git}
-MINICONDA3_VERSION=${MINICONDA3_VERSION:-4.3.21} # Version of Miniconda to install
+# force Python 3
+LSST_PYTHON_VERSION=3
+MINICONDA_VERSION=${MINICONDA_VERSION:-4.3.21}
 MINICONDA_BASE_URL=${MINICONDA_BASE_URL:-https://repo.continuum.io/miniconda}
 CONDA_CHANNELS=${CONDA_CHANNELS:-""}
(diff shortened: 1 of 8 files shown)


jhoblitt:
Rebased after the merge of #220.

jhoblitt:
We have different package lists for osx and linux.  There is no mechanism for setting feature flags on a package in an env file nor is there a way to limit a package to specific platform (as can be done with conda packaging metadata).  This is the functionality that conda provides.  I am unenthusiastic about doing extra mangling rather than using defacto conda env format files.  Would you prefer to adopt `nomkl` for osx?

timj:
We have gone from a single source of truth to having to support two different files with mostly the same content. My suggestion is to merge a generic text file with an OS-specific text file and setting up the environment from the combination. It would be even better if the files could have comment lines in them to indicate which conda packages are needed for the build environment and which are needed for the LSST packages.

jhoblitt:
BTW - I also agree that the deps for `lsst_build` shouldn't even be part of the conda env files but I don't presently have time to convert it to a pypi package.

timj:
Don't you think that duplicating the list of conda packages is going to cause an issue? Ideally we wouldn't need the `nomkl` anomaly but I don't see [DM-8146](https://jira.lsstcorp.org/browse/DM-8146) getting fixed any time soon (@yalsayyad can confirm). As a compromise can you ensure that the files are properly commented so that the `nomkl` entry is associated with DM-8146 and that the lsst_build dependencies are explicitly separated in the list, with documentation that once DM-8146 is closed that we will merge to a single file?

jhoblitt:
Considering the rate of change in the the unpinned conda package list, I'm more concerned about magic than I am two text files falling out of sync.  I much rather that per platform restrictions be added upstream in conda itself.  As this is already supported for conda package build metadata, I suspect it would be an acceptable feature.

jhoblitt:
I thought there was an open jira issue on publishing `lsst_build` to pypy / `pip` installing it but I couldn't find one so I've opened DM-15025.

jhoblitt:
Ugh -- it appears that conda env files only allow `#` as a header and not mixed in between package specifications.

PR #219, opened 2018-07-03
update  conda packages

jhoblitt:
Note that miniconda 4.5.4 is the most recent version and ships with python 3.6.5 but the conda environment files update python to 3.6.6.

based on #218


bin/deploy
```diff
@@ -7,16 +7,16 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
...
-MINICONDA_VERSION=${MINICONDA_VERSION:-4.3.21}
-MINICONDA_BASE_URL=${MINICONDA_BASE_URL:-https://repo.continuum.io/miniconda}
-CONDA_CHANNELS=${CONDA_CHANNELS:-""}
-GIT_VERSION=${GIT_VERSION:-2.16.1}          # Version of git to install
-LFS_VERSION=${LFS_VERSION:-2.3.4}           # Version of git-lfs to install
+LSST_MINICONDA_VERSION=${LSST_MINICONDA_VERSION:-4.5.4}
+LSST_MINICONDA_BASE_URL=${LSST_MINICONDA_BASE_URL:-https://repo.continuum.io/miniconda}
+LSST_CONDA_CHANNELS=${LSST_CONDA_CHANNELS:-""}
+LSST_GIT_VERSION=${LSST_GIT_VERSION:-2.16.1}
+LSST_LFS_VERSION=${LSST_LFS_VERSION:-2.3.4}
```
timj: Is it worth bumping lfs at the same time? I'm using 2.4.2.

bin/deploy
```diff
@@ -7,16 +7,16 @@ SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
...
+LSST_EUPS_GITREV=${LSST_EUPS_GITREV:-""}
+LSST_EUPS_GITREPO=${LSST_EUPS_GITREPO:-https://github.com/RobertLuptonTheGood/eups.git}
 # force Python 3
 LSST_PYTHON_VERSION=3
-MINICONDA_VERSION=${MINICONDA_VERSION:-4.3.21}
-MINICONDA_BASE_URL=${MINICONDA_BASE_URL:-https://repo.continuum.io/miniconda}
-CONDA_CHANNELS=${CONDA_CHANNELS:-""}
-GIT_VERSION=${GIT_VERSION:-2.16.1}          # Version of git to install
-LFS_VERSION=${LFS_VERSION:-2.3.4}           # Version of git-lfs to install
+LSST_MINICONDA_VERSION=${LSST_MINICONDA_VERSION:-4.5.4}
```
timj: Can you take a look at DM-12214 and close it if it is now irrelevant. DM-12265, DM-12266 and DM-12267 need to be looked at as well. They were all listed as blockers for Anaconda 5.0

etc/conda3_packages-linux-64.txt
```diff
@@ -1,66 +1,84 @@
...
-cffi=1.10.0=py36_0
-cryptography=1.8.1=py36_0
-cycler=0.10.0=py36_0
-cython=0.26=py36_0
-dbus=1.10.20=0
-expat=2.1.0=0
-fontconfig=2.12.1=3
-freetype=2.5.5=2
+asn1crypto=0.24.0=py36_0
+astropy=3.0.3=py36h14c3975_2
```
timj: Have you tested this configuration? I'm pretty sure that astropy 3.0.3 breaks everything (DM-14525)

bin/deploy
jhoblitt: Sure. Might as well bump git too even though I'd like to eventually remove it from lsstsw.

jhoblitt:
I haven't run through a full stack build.  Adding the 'main' channel to the jenkins conda mirror finish this afternoon... several packages seem to have stopped being updated in the `free` changed but new versions are in `main` even though the appear to be open source versions without linkage to mkl/etc.  Then then I smacked into our usage of `CONDA_CHANNELS` conflicted with the conda usage of same env var... which was both undocumented and broken... and is now merely undocumented.

timj:
I'm sure the validata_drp problem with astropy is easy to fix, but someone in SQuaRE has to be trying to fix it if we want to update astropy.

etc/conda3_packages-linux-64.txt
jhoblitt: Not even remotely close to running `validate_drp` -- I guess we'll have to put a constraint in the bleed file (maybe bleed is a poor choice of name...).

timj: or have someone fix `validate_drp`...

bin/deploy
jhoblitt: I had forgotten about the RFC and the related tickets.  I'm linking everything to DM-14011 but can't say much until I've made it through a full a test build... hopefully overnight.

jhoblitt:
I've split out and merged all of the non miniconda installer/conda package version improvements as #223.

jhoblitt:
The miniconda installer update has been split out and merged as #225.

etc/conda3_packages-linux-64.txt
timj: I think it was matplotlib that was the problem with DM-14525 and 2.2.2 (which we are using here) fixes the problem.

PR #220, opened 2018-07-03
travis improvements

jhoblitt:
Additional tests, move large shell fragments out of `.travis.yaml`, and switch to running acceptance tests inside of a docker container.  There have been several unexplained failures in the default travis build env with truncated console output and switching over to our own docker image appears to resolve this.  However, it may add up to several minutes to each travis "job".

### DIFF
--- a/.mdlrc
+++ b/.mdlrc
@@ -1,0 +1,5 @@
+# a seperate "style" file must be used to pass "parameters" to a rule
+#
+# https://github.com/markdownlint/markdownlint/blob/master/docs/configuration.md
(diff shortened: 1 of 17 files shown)


.mdlrc
```diff
@@ -0,0 +1,5 @@
+# a seperate "style" file must be used to pass "parameters" to a rule
```
timj: Typo: separate

jhoblitt: Argh.  I bet this exists in several repos from copying this file again.  Fixed.

PR #226, opened 2018-07-09
Bump conda pytest versions to 3.6.2

brianv0:
This will make sure we are in line with the pytest version that we declare in eups


jhoblitt:
(and causes version conflicts)

brianv0:
Closed because of duplication with #219

PR #228, opened 2018-08-29
add cffi to the bleeding edge dependency lists

danielsf:
necessary for GalSim 2.0.6


danielsf:
@jhoblitt DM-15586

timj:
cffi is already in the versioned package list because one of our other dependencies brings it in. No change is needed elsewhere in lsstsw to support this change, it's formalizing cffi.

PR #227, opened 2018-07-26
DM-15142: Create CI-sized DECam Dataset

kfindeisen:
This PR registers all currently existing `ap_verify` datasets as undownloadable by `eups distrib`. All datasets are Git LFS repositories.

PR #229, opened 2018-09-20
add python to bleed conda env(s)

jhoblitt:
This is needed to fix conda downgrading python 3.6 -> 3.5 (even though
3.7 is available) when installing the "bleed" packages.


jhoblitt:
This is needed to resolve the travis failure seen in #227

jhoblitt:
@timj This fix is only required for running under bleed mode, the `LSSTSW_REF` in use shouldn't need to be changed.

PR #230, opened 2018-12-26
Update to use the new scipipe_conda_env repo

gcomoretto:
The changes required to work with the new scipipe_conda_env are done.

No changes have been made in the implemented logic, just the new git package is used.


bin/deploy
```diff
@@ -164,12 +165,30 @@ main() {
...
   local conda_packages
   conda_packages="conda${pyver_prefix}_${deploy_mode}-${pkg_postfix}.txt"
-  local env_file="${SCRIPT_DIR}/../etc/${conda_packages}"
-  local lsstsw_ref
-  lsstsw_ref=$(git log -n 1 --pretty=format:'%h' "$env_file")
+  #local env_file="${SCRIPT_DIR}/../etc/${conda_packages}"
+  #local lsstsw_ref
+  #lsstsw_ref=$(git log -n 1 --pretty=format:'%h' "$env_file")
+  #
+  #echo "::: effective LSST_LSSTSW_REF: ${lsstsw_ref}"
```
jhoblitt: unnecessary commented code.

bin/deploy
```diff
@@ -164,12 +165,30 @@ main() {
...
 
-  echo "::: effective LSST_LSSTSW_REF: ${lsstsw_ref}"
+  echo "::: conda environment file: ${env_file}"
+
+  local env_ref
+  cd env
+  if [ -d 'scipipe_conda_env' ]; then
+    echo ":::  conda environment already present"
+  else 
+    git clone "$LSST_BUILD_GITREPO"
```
jhoblitt: Cloning the repo is is problematic as it will not update if/when `deploy` is re-run.  I think it would more reliable to use the same logic as `newinstall.sh` and fetch the env file by ref.

gcomoretto: The intention is to keep the same logic is implemented now, in order to fulfill the scope of RFC-553. Changing the logic is a different issue.

gcomoretto:
Travis was complaining due to an error in the variable I use for cloning. Now it is fine.
I have remove the old etc/conda3_* files. I was aiming to remove them only once CI are running using the new spipipe_conda_env... but right, no reasons to keep them in the branch.

bin/deploy
jhoblitt: The current usage model, which includes how jenkins is using `lsstsw`, is that the working `lsstsw` clone is pulled and the `./bin/deploy` is re-run to update the science pipelines biuld environment.  As an example, jenkins is doing this as the start of every `stack-os-matrix` build.  Changing this model should involve documenting a new end-user workspace and updating the jenkins job(s) that use `lsstsw`.

bin/deploy
```diff
@@ -164,11 +168,24 @@ main() {
...
-  echo "::: effective LSST_LSSTSW_REF: ${lsstsw_ref}"
+  cd "$LSSTSW"
+  # conda environment reference
+  local env_file="${LSSTSW}/env/${LSST_SPLENV_REF}/${conda_packages}"
+  local env_url="https://raw.githubusercontent.com/lsst/scipipe_conda_env/${LSST_SPLENV_REF}/etc/"
+
+  echo "::: conda environment file: ${env_file}"
+
+  cd env
+  if [ -d "${LSST_SPLENV_REF}" ]; then
```
jhoblitt: It would be better to check the complete local file path as otherwise, trying to switch from a "bleed" deploy to regular (or vise versa), the proper conda env file wouldn't be retrieved.

jhoblitt:
This PR still needs a descriptive title.

gcomoretto:
PR title updated.

PR #234, opened 2019-05-03
DM-18112: Updates to deal with multiple environments

gcomoretto:
bin/deploy
   updated to deploy specific scipipe_conda_env references

bin/rebuild
  updated to store requirement information

bin/envconfig
  new to setup different available environment previously deployed

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -20,6 +20,7 @@ LSST_LFS_VERSION=${LSST_LFS_VERSION:-2.4.2}
 LSST_BUILD_GITREV=${LSST_BUILD_GITREV:-master}
 LSST_BUILD_GITREPO=${LSST_BUILD_GITREPO:-https://github.com/lsst/lsst_build.git}
 LSST_SCIPIPECENV_GITREPO=${LSST_SCIPIPECENV_GITREPO:-https://github.com/lsst/scipipe_conda_env.git}
+LSST_SPLENV_REF=${LSST_SPLENV_REF:-1172c30}
 
 set -e
 
@@ -183,7 +184,6 @@ main() {
     rawreftip=$(git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVREF})
     reftip=$(echo "$rawreftip" | cut -c1-7)
     reftype=$( echo "$rawreftip" | cut -f 2 -d ' ' | cut -f 2 -d'/' )
-    echo ">> $reftype"
     if [ "$reftip" == '' ] || [ "$reftype" == 'tags' ]; then
         # the provided ref is not a branch
         LSST_SPLENV_REF="${ENVREF}"
@@ -194,18 +194,16 @@ main() {
         # the provided ref is a branch
         LSST_SPLENV_REF="$reftip"
         # Defining environment name based on branch and SHA1 from the tip of the branch
-        LSST_CONDA_ENV_NAME="scipipe-env-${FIXED_ENVREF}.${LSST_SPLENV_REF}"
+        LSST_CONDA_ENV_NAME="lsst-scipipe-${FIXED_ENVREF}.${LSST_SPLENV_REF}"
         local env_file="${LSSTSW}/env/${FIXED_ENVREF}/${reftip}/${conda_packages}"
     fi
   else
-    LSST_SPLENV_REF=${LSST_SPLENV_REF:-1172c30}
     local env_file="${LSSTSW}/env/${LSST_SPLENV_REF}/${conda_packages}"
   fi
 
   cd "$LSSTSW"
   # conda environment reference
   local env_url="https://raw.githubusercontent.com/lsst/scipipe_conda_env/${LSST_SPLENV_REF}/etc/"
-  echo "$env_url"
 
   echo "::: conda environment file: ${env_file}"
 
--- a/bin/deploy
+++ b/bin/deploy
@@ -45,13 +45,15 @@ usage() {
   # note that heredocs are prefixed with tab chars
   fail "$(cat <<-EOF
 
-		Usage: $0 [-2] [-3] [-b] [-h]
+		Usage: $0 [-2] [-3] [-b] [-h] [-e] [-r]
 
 		Specific options:
-		    -b          use bleeding edge conda packages
-		    -3          use Python 3 (default)
-		    -2          use Python 2 (no longer supported -- fatal error)
-		    -h          show this message
+        -b          use bleeding edge conda packages
+        -e REF      sperimental env ref
+        -r BRANCH   sperimental branch ref
+        -3          use Python 3 (default)
+        -2          use Python 2 (no longer supported -- fatal error)
+        -h          show this message
 
 		EOF
   )"
@@ -95,7 +97,7 @@ parse_args() {
 
   # We are intentionally not using gnu `getopt` due to portability concerns.
   # Sadly, this means no long options without a massive amount of boilerplate.
-  while getopts "23bh" opt; do
+  while getopts "23bher" opt; do
     case "$opt" in
     b)
       BLEED_DEPLOY=true
@@ -108,6 +110,14 @@ parse_args() {
       ;;
     h)
       usage
+      ;;
+    e)
+      shift;
+      ENVREF=$1
+      ;;
+    r)
+      shift;
+      ENVBREF=$1
       ;;
     *)
       usage "Unknown option: ${opt}"
@@ -132,12 +142,33 @@ am_I_sourced() {
   fi
 }
 
+get_branch_tip_hash() {
+    reftip=`git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVBREF} | cut -c1-7`
+    if [[ $reftip == '' ]]; then
+      echo ":::- Branch ${ENVBREF} invalid!"
+      exit
+    else
+      echo ":::- Found new environment reference ${reftip} !"
+    fi
+}
+
+fixnames() {
+  FIXED_ENVREF=${ENVREF//[.\/]/-}
+  FIXED_ENVBREF=${ENVBREF//[.\/]/-}
+  echo "Fixed ${FIXED_ENVREF}"
+  echo "Fixed ${FIXED_ENVBREF}"
+}
+
 main() {
   config_curl
 
   BLEED_DEPLOY=false
 
+  ENVREF=""
+  ENVBREF=""
+
   parse_args "$@"
+  fixnames
 
   mkdir -p "${LSSTSW}"/{sources,build,var/run,var/log,lfs,distserver/production,etc,env}
 
@@ -166,26 +197,47 @@ main() {
       ;;
   esac
 
+  # If -r is provided, -e is ignored
+  if [[ $FIXED_ENVBREF != '' ]]; then
+    # get given branch SHA1 and assign it to LSST_SPLENV_REF
+    get_branch_tip_hash
+    LSST_SPLENV_REF="$reftip"
+    # Defining environment name
+    LSST_CONDA_ENV_NAME="${LSST_CONDA_ENV_PREFIX}-${FIXED_ENVBREF}-${LSST_SPLENV_REF}"
+  elif [[ $ENVREF != '' ]]; then
+    LSST_SPLENV_REF="${ENVREF}"
+    # Defining environment name
+    LSST_CONDA_ENV_NAME="${LSST_CONDA_ENV_PREFIX}-${FIXED_ENVREF}"
+  fi
+  REF_FOLDER_NAME=${LSST_SPLENV_REF//[.\/]/-}
+
+  echo "Ref Name: ${LSST_SPLENV_REF}"
+  echo "Folder name: ${REF_FOLDER_NAME}"
+  echo "Env Name: ${LSST_CONDA_ENV_NAME}"
+
   local conda_packages
   conda_packages="conda${pyver_prefix}_${deploy_mode}-${pkg_postfix}.yml"
 
   cd "$LSSTSW"
   # conda environment reference
-  local env_file="${LSSTSW}/env/${LSST_SPLENV_REF}/${conda_packages}"
+  local env_file="${LSSTSW}/env/${FIXED_ENVBREF}/${REF_FOLDER_NAME}/${conda_packages}"
   local env_url="https://raw.githubusercontent.com/lsst/scipipe_conda_env/${LSST_SPLENV_REF}/etc/"
 
   echo "::: conda environment file: ${env_file}"
 
   cd env
-  if [ -e "${LSST_SPLENV_REF}"/"${conda_packages}" ]; then
+  if [ -e "${env_file}" ]; then
     echo ":::  conda environment already present"
   else
-    mkdir "${LSST_SPLENV_REF}"
+    if [[ $FIXED_ENVBREF != '' ]]; then
+      mkdir -p "${FIXED_ENVBREF}"
+      cd "${FIXED_ENVBREF}"
+    fi
+    mkdir -p "${REF_FOLDER_NAME}"
     $CURL "${CURL_OPTS[@]}" -# -L \
        "${env_url}/${conda_packages}" \
-       --output "${LSST_SPLENV_REF}/${conda_packages}"
+       --output "${REF_FOLDER_NAME}/${conda_packages}"
   fi
-  echo "::: conda environment revision: ${LSST_SPLENV_REF}"
 
   cd "$LSSTSW"
 
@@ -260,7 +312,8 @@ main() {
 
   # intentionally outside of a lockfile subshell
   # shellcheck disable=SC1091
-  source activate "$LSST_CONDA_ENV_NAME"
+  echo "Activate environment ${LSST_CONDA_ENV_NAME}"
+  source activate "${LSST_CONDA_ENV_NAME}"
 
   # report packages in the current conda env
   #
--- a/bin/deploy
+++ b/bin/deploy
@@ -203,7 +203,7 @@ main() {
     get_branch_tip_hash
     LSST_SPLENV_REF="$reftip"
     # Defining environment name
-    LSST_CONDA_ENV_NAME="${LSST_CONDA_ENV_PREFIX}-${FIXED_ENVBREF}-${LSST_SPLENV_REF}"
+    LSST_CONDA_ENV_NAME="${LSST_CONDA_ENV_PREFIX}-${FIXED_ENVBREF}.${LSST_SPLENV_REF}"
   elif [[ $ENVREF != '' ]]; then
     LSST_SPLENV_REF="${ENVREF}"
     # Defining environment name
--- a/bin/deploy
+++ b/bin/deploy
@@ -20,9 +20,6 @@ LSST_LFS_VERSION=${LSST_LFS_VERSION:-2.4.2}
 LSST_BUILD_GITREV=${LSST_BUILD_GITREV:-master}
 LSST_BUILD_GITREPO=${LSST_BUILD_GITREPO:-https://github.com/lsst/lsst_build.git}
 LSST_SCIPIPECENV_GITREPO=${LSST_SCIPIPECENV_GITREPO:-https://github.com/lsst/scipipe_conda_env.git}
-# this git ref controls which set of conda packages are used to initialize the
-# the default conda env defined in scipipe_conda_env git package (RFC-553).
-LSST_SPLENV_REF=${LSST_SPLENV_REF:-1172c30}
 
 set -e
 
@@ -45,15 +42,14 @@ usage() {
   # note that heredocs are prefixed with tab chars
   fail "$(cat <<-EOF
 
-		Usage: $0 [-2] [-3] [-b] [-h] [-e] [-r]
+		Usage: $0 [-2] [-3] [-b] [-h] [-r]
 
 		Specific options:
-        -b          use bleeding edge conda packages
-        -e REF      github reference in scipipe_conda_env (tag or SHA1)
-        -r BRANCH   github branch in scipipe_conda_env
-        -3          use Python 3 (default)
-        -2          use Python 2 (no longer supported -- fatal error)
-        -h          show this message
+	    -b          use bleeding edge conda packages
+	    -r REF      git ref in scipipe_conda_env (hash, branch, tag)
+	    -3          use Python 3 (default)
+	    -2          use Python 2 (no longer supported -- fatal error)
+	    -h          show this message
 
 		EOF
   )"
@@ -97,7 +93,7 @@ parse_args() {
 
   # We are intentionally not using gnu `getopt` due to portability concerns.
   # Sadly, this means no long options without a massive amount of boilerplate.
-  while getopts "23bher" opt; do
+  while getopts "23bhr" opt; do
     case "$opt" in
     b)
       BLEED_DEPLOY=true
@@ -111,13 +107,9 @@ parse_args() {
     h)
       usage
       ;;
-    e)
-      shift;
-      ENVREF=$1
-      ;;
     r)
       shift;
-      ENVBREF=$1
+      ENVREF=$1
       ;;
     *)
       usage "Unknown option: ${opt}"
@@ -142,20 +134,7 @@ am_I_sourced() {
   fi
 }
 
-get_branch_tip_hash() {
-    reftip=`git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVBREF} | cut -c1-7`
-    if [[ $reftip == '' ]]; then
-      echo ":::- Branch ${ENVBREF} invalid!"
-      exit
-    else
-      echo "::: Found new environment reference ${reftip} !"
-    fi
-}
 
-fixnames() {
-  FIXED_ENVREF=${ENVREF//[.\/]/-}
-  FIXED_ENVBREF=${ENVBREF//[.\/]/-}
-}
 
 main() {
   config_curl
@@ -163,10 +142,10 @@ main() {
   BLEED_DEPLOY=false
 
   ENVREF=""
-  ENVBREF=""
 
   parse_args "$@"
-  fixnames
+  # Create safe directory name from git ref
+  FIXED_ENVREF=${ENVREF//[.\/]/-}
 
   mkdir -p "${LSSTSW}"/{sources,build,var/run,var/log,lfs,distserver/production,etc,env}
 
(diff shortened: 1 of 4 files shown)


bin/rebuild
```diff
@@ -31,6 +32,38 @@ versiondb_head() {
...
+    else
+      local_env_dir="${LSSTSW}/env/${ref}"
+    fi 
+    if [ -d "${local_env_dir}" ]; then
+      ref_file=`ls "${local_env_dir}"`
+      local_env_file="${local_env_dir}/${ref_file}"
+      echo "env_ref_file: ${local_env_file}" >> "${filename}"
+      cat "${local_env_file}" | grep -v "^name" > /tmp/renv
+      set +e  # disable set -e for the diff commands.
+      sdiff /tmp/lenv1 /tmp/renv >> "${filename}"
```
ktlim: An `sdiff` seems hard to use by any other automation.  Why not use the `diff -u` here, if anything is needed at all?

bin/rebuild
```diff
@@ -31,6 +32,38 @@ versiondb_head() {
...
+    ref="${LSST_CONDA_ENV_NAME:12}"
+    if [[ $ref == *"."* ]]; then
+      ref1=`echo "${ref}" | cut -f 1 -d .`
+      ref2=`echo "${ref}" | cut -f 2 -d .`
+      local_env_dir="${LSSTSW}/env/${ref1}/${ref2}"
+    else
+      local_env_dir="${LSSTSW}/env/${ref}"
+    fi 
+    if [ -d "${local_env_dir}" ]; then
+      ref_file=`ls "${local_env_dir}"`
```
ktlim: This seems dangerous.  You should check explicitly for a `conda*.yml` file, not just blindly assume any file (and what if there's zero or more than one?) in the env directory is appropriate.

bin/rebuild
```diff
@@ -31,6 +32,38 @@ versiondb_head() {
   (cd "$VERSIONDB" && git rev-parse HEAD)
 }
 
+save_build_env() {
+  # saving environment information
+  filename="${LSSTSW_BUILD_DIR}/${BUILD}.env"
+  echo "Saving environment information in ${filename}"
+  conda env export | grep -v "^name"> /tmp/lenv
```
ktlim: I think use of the `diff -I` option might avoid a lot of this file manipulation.

bin/rebuild
```diff
@@ -31,6 +32,38 @@ versiondb_head() {
...
+  echo "environment_name: ${LSST_CONDA_ENV_NAME}" > "${filename}"
+  cat /tmp/lenv | grep "^prefix" >> "${filename}"
+  cat /tmp/lenv | grep -v "^prefix" > /tmp/lenv1
+  if [[ "${LSST_CONDA_ENV_NAME}" != "lsst-scipipe" ]]; then
```
ktlim: Why don't we record the environment for `lsst-scipipe`?

bin/deploy
```diff
@@ -255,12 +286,14 @@ main() {
...
 
+    echo "${ARGS[@]}"
     conda "${ARGS[@]}"
   )
 
   # intentionally outside of a lockfile subshell
   # shellcheck disable=SC1091
-  source activate "$LSST_CONDA_ENV_NAME"
+  echo "Activate environment ${LSST_CONDA_ENV_NAME}"
+  source activate "${LSST_CONDA_ENV_NAME}"
```
ktlim: Should we be considering changing to the official `conda activate` mechanism instead of this? https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+INTERACTIVE=false
+ENVREF=""
+BUILD_ID=""
+
+parse_args "$@"
+
+if [[ !$finish ]]; then
+
+
+if [[ -z $ZSH_NAME ]]; then
```
ktlim: This repeats a lot of `setup.sh`.  Can the common code be factored out?

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+export PATH="$LSSTSW/bin:$PATH"
+
+export MANPATH="$LSSTSW/lfs/share/man:"
+
+# need to activate the miniconda base environment to get the correct list of environments
+source activate "${LSSTSW}/miniconda"
+
+if $INTERACTIVE ; then
+  # get the list of available environments
+  #ENVS="`conda env list | grep -v ^# | awk -F ' /' '{ print $2 }'`"
```
ktlim: Remove prototype/debugging code.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+  IFS=$'\n' raw_envs=( $( conda env list | grep -v "^#" | grep -v "^ " | grep -v "^base" ) )
+  envs=()
+  #echo "${ENVS}"
```
ktlim: Remove debugging code.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+fi
+
+export PATH="$LSSTSW/miniconda/bin:$PATH"
+export PATH="$LSSTSW/lfs/bin:$PATH"
+export PATH="$LSSTSW/bin:$PATH"
+
+export MANPATH="$LSSTSW/lfs/share/man:"
+
+# need to activate the miniconda base environment to get the correct list of environments
+source activate "${LSSTSW}/miniconda"
```
ktlim: I guess `base` can't be used here?

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+if $INTERACTIVE ; then
+  # get the list of available environments
+  #ENVS="`conda env list | grep -v ^# | awk -F ' /' '{ print $2 }'`"
+  IFS=$'\n' raw_envs=( $( conda env list | grep -v "^#" | grep -v "^ " | grep -v "^base" ) )
+  envs=()
+  #echo "${ENVS}"
+  l=0
+  for i in "${raw_envs[@]}"; do
+    ((l++))
+    envs[l]=`echo "${i}" | cut -f 1 -d ' '`
```
ktlim: If you did the `cut` in the `conda env list` pipeline, you wouldn't have to have a separate list variable.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+    echo "${l} - ${envs[${l}]}"
+  done
+  read -p "Choose environment [1 to ${l}]: " eid
+  LSST_CONDA_ENV_NAME="${envs[${eid}]}"
```
ktlim: Check to make sure it's in range?

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+  source activate "$LSST_CONDA_ENV_NAME"
+
+elif [[ $ENVREF != "" ]]; then
+  echo "Setting $ENVREF environment..."  
+  LSST_CONDA_ENV_NAME=${ENVREF}
+  source activate "$LSST_CONDA_ENV_NAME"
+
+elif [[ $BUILD_ID != "" ]]; then
+  echo "Looking for evironment used in build ${BUILD_ID}..."
```
ktlim: Typo.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+  if [[ -f "build/${BUILD_ID}.env" ]]; then
```
ktlim: This is only local builds, not ones done by e.g. Jenkins?

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+    LSST_CONDA_ENV_NAME=`cat "build/${BUILD_ID}.env" | grep 'environment_name' | cut -f 2 -d ' '`
+    source activate "$LSST_CONDA_ENV_NAME"
+    echo "Environment ${LSST_CONDA_ENV_NAME} found."
```
ktlim: Be more consistent with messaging ("Activat[ing]" or "Setting" above and below, but just "found" here).

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+export PATH="$LSSTSW/miniconda/envs/$LSST_CONDA_ENV_NAME/bin:$PATH"
+#echo "${PATH}"
+
+. "$LSSTSW/eups/current/bin/setups.$SUFFIX"
+
+setup -r "$LSSTSW/lsst_build"
+
+echo "notice: lsstsw tools have been set up."
+
+# vim: tabstop=2 shiftwidth=2 expandtab
```
ktlim: I think this is supposed to be the last line.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+fi
```
ktlim: And this should go away anyway with use of `exit`.

bin/deploy
```diff
@@ -45,13 +42,14 @@ usage() {
...
 
-		Usage: $0 [-2] [-3] [-b] [-h]
+		Usage: $0 [-2] [-3] [-b] [-h] [-r]
 
 		Specific options:
-		    -b          use bleeding edge conda packages
-		    -3          use Python 3 (default)
-		    -2          use Python 2 (no longer supported -- fatal error)
-		    -h          show this message
+        -b          use bleeding edge conda packages
```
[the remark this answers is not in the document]
gcomoretto: Git diff from command line doesn't show any differences, difficult to see tabs. I change as you suggest: tab plus 4 spaces

bin/deploy
```diff
@@ -45,13 +42,14 @@ usage() {
...
+        -r REF      github sha1/branch in scipipe_conda_env
```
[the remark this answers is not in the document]
gcomoretto: Do you mean `Git Ref` instead of `REF`?

bin/deploy
```diff
@@ -45,13 +42,14 @@ usage() {
...
+        -3          use Python 3 (default)
```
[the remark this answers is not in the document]
gcomoretto: I would prefer to keep this on a different branch. I think the issue is already open, at least to remove `-2`

bin/deploy
```diff
@@ -132,12 +134,17 @@ am_I_sourced() {
...
+
 main() {
   config_curl
 
   BLEED_DEPLOY=false
 
+  ENVREF=""
+
   parse_args "$@"
+  FIXED_ENVREF=${ENVREF//[.\/]/-}
```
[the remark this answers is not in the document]
gcomoretto: The reference provided may be a branch with slashes. I am using that reference to create a folder where to store the corresponding environment file. This is just to try to have the folder name cleaner in the filesystem. It may be necessary to add more substitutions.

bin/deploy
```diff
@@ -169,23 +176,47 @@ main() {
...
+    if [[ $reftip == '' ]]; then
+        # the provided ref is not a branch nor a tag
+        LSST_SPLENV_REF="${ENVREF}"
+        # Defining environment name based on the provided SHA1
+        LSST_CONDA_ENV_NAME="scipipe-env-${FIXED_ENVREF}"
+        local env_file="${LSSTSW}/env/${ENVREF}/${conda_packages}"
+    else
+        LSST_SPLENV_REF="$reftip"
+        # Defining environment name based on branch and SHA1 from the tip of the branch
+        LSST_CONDA_ENV_NAME="scipipe-env-${FIXED_ENVREF}.${LSST_SPLENV_REF}"
```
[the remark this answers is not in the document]
gcomoretto: In this way, you will know more precisely which environment you are using.

You mean, only to store one environment file and overwrite it?
In this way, we have a local copy of the environment file ready to use from the command line, for debug or manual comparison with other environments.

bin/deploy
```diff
@@ -169,23 +176,47 @@ main() {
...
-    mkdir "${LSST_SPLENV_REF}"
+    if [[ $reftip != '' ]]; then
+      # if a branch or tag is provided, store the environment yaml inside the corresponding subfolder
+      mkdir -p "${FIXED_ENVREF}"
+      cd "${FIXED_ENVREF}"
+    fi
+    mkdir -p "${LSST_SPLENV_REF}"
     $CURL "${CURL_OPTS[@]}" -# -L \
        "${env_url}/${conda_packages}" \
        --output "${LSST_SPLENV_REF}/${conda_packages}"
```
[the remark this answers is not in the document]
gcomoretto: This was like this before. I agree that it should be `${env_file}`.

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+		Specific options:
+        -i          interactive, propose a list of available environment
+        -r REF      environment name
+        -b bXXXX    activate the environment already used for the build with id bXXXX
+        -h          show this message
+
+		EOF
+  )"
+
+  finish=true
```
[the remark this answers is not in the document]
gcomoretto: If I use `exit`, that will ternimate the active shell also

bin/envconfig
```diff
@@ -0,0 +1,147 @@
...
+   
+}
+
+parse_args() {
+  local OPTIND
+  local opt
+
+  while getopts "hrib" opt; do
```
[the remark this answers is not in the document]
gcomoretto: OK, I have tried to order them: first: brih

bin/rebuild
```diff
@@ -65,7 +98,8 @@ fi
 
 (
   # shellcheck disable=SC1091
-  source activate "$LSST_CONDA_ENV_NAME"
+  #source activate "$LSST_CONDA_ENV_NAME"
+  # removes since it is not clear why this source is done, since setup.sh shall have already done it
```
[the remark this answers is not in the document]
gcomoretto: And have just that `$LSST_CONDA_ENV_NAME` defined manually?
The user may miss in that way the other configurations that the setup.sh provides.
The probability that something will not work is increasing.

We can let it in, so to keep compatibility with the current way to work.
But this is something that does not help the environment management.

bin/envconfig
gcomoretto: I wanted something that introduces new functionalities and has no `setup` word in its name.
At the same time, I wanted to keep compatibility with the current way to work, so in my opinion, it is important to keep `setup.sh` as it.
If one day everybody is happy with `envconfig`, we can remove `setup.sh` from the tooling.

bin/deploy
gcomoretto: I think this is part of a bigger picture, like using a new conda distribution.

bin/envconfig
gcomoretto: Like above, I think this is part of a bigger picture, like using a new conda distribution.

bin/envconfig
gcomoretto: The logic to change the environment should be managed in Jenkins itself. And the same can be used in case of a re-run.

bin/envconfig
gcomoretto: That line was already there. I can remove it.

bin/deploy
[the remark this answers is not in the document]
ktlim: I meant `-r REF         git ref in scipipe_conda_env`.

bin/deploy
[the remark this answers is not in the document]
ktlim: OK, but could you add a comment like `# Create safe directory name from git ref`?

bin/deploy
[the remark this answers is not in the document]
ktlim: No, I don't mean to only have one environment.  The `LSST_SPLENV_REF` is already unique, though, so I wasn't sure why you needed to add `FIXED_ENVREF` to it.

bin/envconfig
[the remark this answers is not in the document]
ktlim: Sorry, I forgot this is to be `source`d.  Use `return` then.

bin/envconfig
ktlim: I meant that a user cannot easily instantiate the environment from a (known) Jenkins build.

bin/envconfig
ktlim: I meant it's supposed to be the last line of the file to give instructions to the editor.  You added a line after it.

bin/envconfig
gcomoretto: As we commented yesterday, I don't think that Jenkins build will use these functionalities, at least not in a first moment.

bin/rebuild
gcomoretto: The environment name (lsst-scipipe) does not contain any information on the git reference used to create it.
Probably it is not difficult to make this information persistent when the environment is created, maybe just a link. But, in any case, I hope that this use case is dismissed as soon as possible.
Moreover, keeping the nominal behavior in line with what is available now, this should not be required.

bin/rebuild
```diff
@@ -31,6 +32,38 @@ versiondb_head() {
   (cd "$VERSIONDB" && git rev-parse HEAD)
 }
 
+save_build_env() {
+  # saving environment information
+  filename="${LSSTSW_BUILD_DIR}/${BUILD}.env"
+  echo "Saving environment information in ${filename}"
+  conda env export | grep -v "^name"> /tmp/lenv
+  echo "environment_name: ${LSST_CONDA_ENV_NAME}" > "${filename}"
```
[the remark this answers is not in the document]
gcomoretto: This is required in order to get `envconfig` configuring the same enviroment.

bin/envconfig
gcomoretto: I prefer to let it here for the moment.

bin/rebuild
gcomoretto: So far I prefer to keep it as is.

gcomoretto:
So far I should have implemented all requested changes.  A few comments have not been taken in account, in order to keep the coding simpler.
Also, all shellcheck failures have been fixed.

I strongly recommend, before merging, somebody else than me to try to use the new code, at least to ensure that it is at least still working as it is used now. @timj  maybe?

bin/deploy
```diff
@@ -45,13 +42,14 @@ usage() {
...
 		Specific options:
-		    -b          use bleeding edge conda packages
-		    -3          use Python 3 (default)
-		    -2          use Python 2 (no longer supported -- fatal error)
-		    -h          show this message
+	    -b          use bleeding edge conda packages
+	    -r REF      git ref in scipipe_conda_env (hash, branch, tag)
+	    -3          use Python 3 (default)
+	    -2          use Python 2 (no longer supported -- fatal error)
+	    -h          show this message
```
timj: Minor comment, but why did the tabbing change here? It makes the diff bigger than it really is (which was adding a `-r` option.

bin/deploy
```diff
@@ -169,23 +177,52 @@ main() {
   local conda_packages
   conda_packages="conda${pyver_prefix}_${deploy_mode}-${pkg_postfix}.yml"
 
+  if [[ $FIXED_ENVREF != '' ]]; then
+    # (try to) get latest branch SHA1 and assign it to LSST_SPLENV_REF
+    # if the REF is not a branch (or a tag), the reftip will be empty
+    rawreftip=$(git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVREF})
+    reftip=$(echo "$rawreftip" | cut -c1-7)
+    reftype=$( echo "$rawreftip" | cut -f 2 -d ' ' | cut -f 2 -d'/' )
+    echo ">> $reftype"
```
timj: I think this debug line should be removed or else be some informative message saying something about the default conda env being installed.

bin/deploy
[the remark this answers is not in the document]
timj: `FIXED_ENVREF` is a mangled version of the supplied `-r` argument?  The logic is therefore that if `-r` is a tag name then you'll use that tag name directly, if it's a branch name you are using the name of the branch and the explicit sha1 associated with that branch? This is so that when someone reports what env they are using they indicate that it came from a branch request? Seems okay to me

bin/deploy
```diff
@@ -169,23 +177,52 @@ main() {
...
+    if [ "$reftip" == '' ] || [ "$reftype" == 'tags' ]; then
+        # the provided ref is not a branch
+        LSST_SPLENV_REF="${ENVREF}"
+        # Defining environment name based on the provided SHA1
+        LSST_CONDA_ENV_NAME="lsst-scipipe-${FIXED_ENVREF}"
```
timj: The env name here starts with `lsst-` but in line 197 it doesn't

bin/deploy
```diff
@@ -169,23 +177,52 @@ main() {
...
+        local env_file="${LSSTSW}/env/${ENVREF}/${conda_packages}"
+    else
+        # the provided ref is a branch
+        LSST_SPLENV_REF="$reftip"
+        # Defining environment name based on branch and SHA1 from the tip of the branch
+        LSST_CONDA_ENV_NAME="scipipe-env-${FIXED_ENVREF}.${LSST_SPLENV_REF}"
+        local env_file="${LSSTSW}/env/${FIXED_ENVREF}/${reftip}/${conda_packages}"
+    fi
+  else
+    LSST_SPLENV_REF=${LSST_SPLENV_REF:-1172c30}
```
timj: Moving the default env SHA1 from line 25 to line 201 makes it much much harder to notice.  I think you can move this line back to line 25 and the logic will still work properly won't it?

bin/deploy
```diff
@@ -169,23 +177,52 @@ main() {
...
+    local env_file="${LSSTSW}/env/${LSST_SPLENV_REF}/${conda_packages}"
+  fi
+
   cd "$LSSTSW"
   # conda environment reference
-  local env_file="${LSSTSW}/env/${LSST_SPLENV_REF}/${conda_packages}"
   local env_url="https://raw.githubusercontent.com/lsst/scipipe_conda_env/${LSST_SPLENV_REF}/etc/"
+  echo "$env_url"
```
timj: debug line?

bin/deploy
```diff
@@ -45,13 +42,14 @@ usage() {
...
 
-		Usage: $0 [-2] [-3] [-b] [-h]
+		Usage: $0 [-2] [-3] [-b] [-h] [-r]
 
 		Specific options:
-		    -b          use bleeding edge conda packages
-		    -3          use Python 3 (default)
-		    -2          use Python 2 (no longer supported -- fatal error)
-		    -h          show this message
+	    -b          use bleeding edge conda packages
```
timj: We might want to tweak the message here. `-b` means that the conda package list as defined in the `-r REF` (or default) is used without version constraints.  That list of packages can be different depending on the choice of `-r`.
How about:

> install conda packages without version constraints

?

bin/deploy
timj: We should definitely update our conda management (and support the new conda 4.6) but I'm happy for that to be a separate ticket (although if the change is only to change "source" to "conda" then we should do it rather than talking about it).

bin/envconfig
timj: to follow up, why can't this be `conda activate base` ?

bin/envconfig
```diff
@@ -0,0 +1,163 @@
+#!/bin/bash
+
+# setup lsstsw environment
```
timj: Are people required to use this script? Or can they use `conda env list` themselves and `conda activate`?

bin/rebuild
timj: I'd be inclined to make every conda env, including the default one that bin/deplot creates, use the new naming convention.

bin/rebuild
```diff
@@ -31,6 +32,39 @@ versiondb_head() {
   (cd "$VERSIONDB" && git rev-parse HEAD)
 }
 
+save_build_env() {
+  # saving environment information
+  filename="${LSSTSW_BUILD_DIR}/${BUILD}.env"
```
timj: These need to go in a subdirectory otherwise the `build/` directory is going to be extremely cluttered. I've run two tests and already feeling like this is getting out of control.

bin/deploy
gcomoretto: I don't know... I did not change the original formatting, so I just fixed the tab problem as KT pointed out.

bin/deploy
gcomoretto: line 197 fixed.

bin/deploy
gcomoretto: Fine for me. 
Maybe `Install environment using bleed packages`.
However, if we want to improve this aspected now, some logic should be added to have a different environment name in this case. This will make obsolete DM-17823.

bin/envconfig
gcomoretto: Other environment variables are defined here, not only the conda environment to use.
The user may have already defined those in the .bash_profile.
However, if the scope of this tool is to provide a ready to use environment, I think the user is required to use this script.

bin/rebuild
gcomoretto: OK, what about `build/builds/`?

bin/rebuild
gcomoretto: If we agree on this, no problems from my side, however, we need to change DMTN-110 accordingly.

bin/deploy
gcomoretto: I prefer a separate ticket.

bin/deploy
gcomoretto: Moved back to line 23

bin/envconfig
gcomoretto: `bin/deploy` is deploying the environments under the local miniconda. If we use `base` conda, we will not be able to see the miniconda environments.
We can add extra logic in order to do that but, in my opinion, it is not in the scope of this issue, and will complicate the scripting quite a lot, increasing the possibility of errors.
In other words, I prefer to keep it simple, at least for the moment.

gcomoretto:
Thank you, Tim, for your comment.
I think the conda upgrade stuff is a big topic and deserves its own ticket.

If we all agree to change the naming convention of the default environment, we can update DMTN-110 in order to be consistent.

I pushed a few changes that should address the latest comments.

bin/envconfig
timj: I'm not sure I follow. If I do:
```
$ source activate `pwd`/miniconda
```
in my shell then I see that my activated environment becomes `base`.  It therefore seems much clearer for us to do `conda activate base` directly.  The only way these might be different is if somehow the path to `conda` is not correct.

gcomoretto:
@ktlim, I think that all requested changes have been implemented.

Is there something else to do before merging this into master?
Maybe check with Josh or Jim?

bin/deploy
ktlim: I think your editor may be reformatting this?  You need to make this a one-line diff somehow, with that one line matching the formatting of the others when this is executed.

bin/deploy
```diff
@@ -169,23 +178,49 @@ main() {
...
+    # if the REF is not a branch (or a tag), the reftip will be empty
+    rawreftip=$(git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVREF})
+    reftip=$(echo "$rawreftip" | cut -c1-7)
+    reftype=$( echo "$rawreftip" | cut -f 2 -d ' ' | cut -f 2 -d'/' )
+    if [ "$reftip" == '' ] || [ "$reftype" == 'tags' ]; then
+        # the provided ref is not a branch
+        LSST_SPLENV_REF="${ENVREF}"
+        # Defining environment name based on the provided SHA1
+        LSST_CONDA_ENV_NAME="lsst-scipipe-${FIXED_ENVREF}"
+        local env_file="${LSSTSW}/env/${ENVREF}/${conda_packages}"
```
ktlim: I think this should be `${FIXED_ENVREF}`.

bin/deploy
```diff
@@ -169,23 +178,49 @@ main() {
...
   echo "::: conda environment file: ${env_file}"
 
   cd env
-  if [ -e "${LSST_SPLENV_REF}"/"${conda_packages}" ]; then
-    echo ":::  conda environment already present"
+  if [ -e "${env_file}" ]; then
+    echo "::: conda environment already present"
   else
-    mkdir "${LSST_SPLENV_REF}"
+    if [[ $reftip != '' ]]; then
```
ktlim: I think this whole if can be replaced by `mkdir -p ${env_file%/*}`

bin/deploy
```diff
@@ -169,23 +178,49 @@ main() {
   local conda_packages
   conda_packages="conda${pyver_prefix}_${deploy_mode}-${pkg_postfix}.yml"
 
+  if [[ $FIXED_ENVREF != '' ]]; then
+    # (try to) get latest branch SHA1 and assign it to LSST_SPLENV_REF
+    # if the REF is not a branch (or a tag), the reftip will be empty
+    rawreftip=$(git ls-remote https://github.com/lsst/scipipe_conda_env.git ${ENVREF})
+    reftip=$(echo "$rawreftip" | cut -c1-7)
+    reftype=$( echo "$rawreftip" | cut -f 2 -d ' ' | cut -f 2 -d'/' )
```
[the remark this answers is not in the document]
gcomoretto: OK, I go for one line improved solution

gcomoretto:
I have added a few improvements, as suggested by @ktlim

PR #237, opened 2019-08-13
DM-20972: Update miniconda to 4.7.10

gcomoretto:
These changes should be sufficient to support @brianv0 new environment.
I have also added a couple of improvements in the messaging.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -455,7 +455,9 @@ main() {
     )
   fi
 
-  echo "Done. Run the following:"
+  echo ""
+  echo "::: Deploy completed."
+  echo "::: To active the just deployed environment use the following command:"
   echo
 
   if [[ $ENVREF == '' ]]; then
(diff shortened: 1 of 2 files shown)


bin/deploy
```diff
@@ -455,7 +455,9 @@ main() {
     )
   fi
 
-  echo "Done. Run the following:"
+  echo ""
+  echo "::: Deploy completed."
+  echo "::: To active the just deployed environment use the following command:"
```
timj: "To activate..."

bin/deploy
```diff
@@ -455,7 +455,9 @@ main() {
     )
   fi
 
-  echo "Done. Run the following:"
+  echo ""
+  echo "::: Deploy completed."
```
timj: I don't really understand what the `::: ` is for.

bin/deploy
```diff
@@ -466,17 +468,16 @@ main() {
     elif [[ $SHELL =~ zsh$ ]]; then
       echo "    . ${LSSTSW}/bin/setup.sh"
     else
-      echo "    The setup script in ${LSSTSW}/bin most appropriate for ${SHELL}"
-      echo "    (Warning: ${SHELL} is an unsupported shell)"
+      echo "::: The setup script in ${LSSTSW}/bin most appropriate for ${SHELL}"
```
timj: why the change from spaces to colons?

bin/envconfig
```diff
@@ -114,7 +114,11 @@ if $INTERACTIVE ; then
     envs[l]=$(echo "${i}" | cut -f 1 -d ' ')
     echo "${l} - ${envs[${l}]}"
   done
+  echo "0 EXIT (return)"
   read -r -p "Choose environment [1 to ${l}]: " eid
```
timj: Maybe include the `0 to exit` in the prompt?

bin/deploy
gcomoretto: I tried to follow what was done above, but probably I should have not changed other's print outs.

bin/deploy
gcomoretto: same as above. I Put back the spaces.

bin/envconfig
gcomoretto: 0 is not an environment. But if you feel it is better to include it, I do.

timj: I mean a prompt something like:
```
Choose environment [1 to ${l}; 0 to exit]: 
```

PR #248, opened 2020-03-27
Create different stack folder for each environment

gcomoretto:
I have done some tests and it seems it works.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -359,6 +359,11 @@ main() {
     touch "${LSSTSW}/lfs/.git-lfs.deployed"
   )
 
+  # definition EUPS_PATH depending on the environment:
+  # including the last 7 characters of the environment name
+  export EUPS_PATH="${LSSTSW}/stack${LSST_CONDA_ENV_NAME: -7}"
+  echo "EUPS_PATH defined to ${EUPS_PATH}"
+
   # backwards compatibility if EUPS wasn't installed to a versioned directory
   test -f "${LSSTSW}/eups/.deployed" && ( # EUPS
     echo "::: Moving old EUPS to eups/legacy"
@@ -396,7 +401,7 @@ main() {
     ./configure \
       --prefix="${LSSTSW}/eups/${LSST_EUPS_VERSION}" \
       --with-python="${LSSTSW}/miniconda/bin/python" \
-      --with-eups="${LSSTSW}/stack"
+      --with-eups="${EUPS_PATH}"
     make
     make install
     touch "${LSSTSW}/eups/${LSST_EUPS_VERSION}/.deployed"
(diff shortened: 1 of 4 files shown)


bin/envconfig.csh
```diff
@@ -137,6 +137,12 @@ conda activate "$LSST_CONDA_ENV_NAME"
 
 source "$LSSTSW/eups/current/bin/setups.csh"
 
+# definition EUPS_PATH depending on the environment:
+# including the last 7 characters of the environment name
+set env_ref = `echo $LSST_CONDA_ENV_NAME | awk '{print substr($0, length($0) - 6)}'`
+set EUPS_PATH="$LSSTSW/stack$env_ref"
```
ktlim, line 143: As above, replace this with `set EUPS_PATH="$LSSTSW/stack-$LSST_SPLENV_REF"`?

bin/deploy
```diff
@@ -359,6 +359,11 @@ main() {
     touch "${LSSTSW}/lfs/.git-lfs.deployed"
   )
 
+  # definition EUPS_PATH depending on the environment:
+  # including the last 7 characters of the environment name
+  export EUPS_PATH="${LSSTSW}/stack${LSST_CONDA_ENV_NAME: -7}"
```
ktlim, line 364: It may be better to remove the prefix rather than count characters in the suffix, but that part is OK.  If you do keep that, then you should probably make the same change to my code in `publish`.

I think it's better to set off the reference hash from the name of the stack by using `/stack-` rather than just `/stack`.

bin/deploy
```diff
@@ -359,6 +359,11 @@ main() {
...
+  echo "EUPS_PATH defined to ${EUPS_PATH}"
```
ktlim, line 365: "set to" may be better than "defined to", here and in all other places.

bin/envconfig
```diff
@@ -164,6 +164,11 @@ export LSST_CONDA_ENV_NAME
 # shellcheck disable=SC1090
 . "$LSSTSW/eups/current/bin/setups.$SUFFIX"
 
+# definition EUPS_PATH depending on the environment:
+# including the last 7 characters of the environment name
+export EUPS_PATH="${LSSTSW}/stack${LSST_CONDA_ENV_NAME: -7}"
```
ktlim, line 169: Can't you just reuse `${LSST_SPLENV_REF}` here rather than computing it?  And please add a `-`.

bin/deploy
```diff
@@ -359,6 +359,11 @@ main() {
     touch "${LSSTSW}/lfs/.git-lfs.deployed"
   )
 
+  # definition EUPS_PATH depending on the environment:
```
ktlim, line 362:
```suggestion
  # Define EUPS_PATH depending on the environment,
```

ktlim: This goes for the other places as well.

bin/envconfig.csh
gcomoretto: Better using subfolders

bin/deploy
gcomoretto: Better using subfolders.
LSST_SPLENV_REF is available, so we can simplify.

bin/envconfig
gcomoretto: Sorry, I miss this point. envconfig will deploy whatever environment, and, from the way the envnames are formatted, the LSSP_SPLENV_REF is the last part of the name. I can't use the $LSST_SPLENV_REF here since, this is fixed to the default one, but envconfig shall be able to activate whatever environment, and assign the proper EUPS_PATH, not only the default one.

PR #258, opened 2020-05-21
DM-25017: Simplify and correct ref argument handling.

gcomoretto:


bin/deploy
```diff
@@ -334,29 +334,34 @@ main() {
...
-
-    case $(uname -s) in
-      Linux*)  lfs_platform="linux-amd64" ;;
-      Darwin*) lfs_platform="darwin-amd64" ;;
-      *)
-        fail "Cannot install git-lfs: unsupported platform $(uname -s)"
-        ;;
-    esac
+  # check if git-lfs is available (in scipipe_conda_env since w_2020_19)
+  if (which git-lfs >/dev/null 2>&1); then
```
ktlim, line 338: You wouldn't even have to change the indent if you added this as an extra `||` test before the existing one, but this is fine (and allows for the extra message).

PR #259, opened 2020-05-26
DM-24972: Support conda lock files

brianv0:
This supports conda lock files (but keeps support for bleed)

It also reorders where `conda config` is ran so it's ran in the environment properly.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -11,7 +11,6 @@ LSST_EUPS_VERSION=${LSST_EUPS_VERSION:-2.1.5}
 LSST_EUPS_GITREV=${LSST_EUPS_GITREV:-""}
 LSST_EUPS_GITREPO=${LSST_EUPS_GITREPO:-https://github.com/RobertLuptonTheGood/eups.git}
 # force Python 3
-LSST_PYTHON_VERSION=3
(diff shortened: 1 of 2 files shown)


bin/deploy
```diff
@@ -273,20 +264,30 @@ main() {
     # installation attempt.  These need to be cleaned up before [re]attempting
     # to install packages.
     conda clean -y --all
+    if [[ $deploy_mode == "bleed" ]]; then
+      ARGS=()
+      ARGS+=('env' 'update')
+      ARGS+=('--name' "$LSST_CONDA_ENV_NAME")
+
+      # disable the conda install progress bar when not attached to a tty. Eg.,
```
ktlim, line 272: This comment is in the wrong place; it belongs next to `--quiet`.

bin/deploy
```diff
@@ -295,6 +296,23 @@ main() {
   # shellcheck disable=SC1091
   conda activate "${LSST_CONDA_ENV_NAME}"
 
+  (
+    # configure alt conda channel(s)
+    if [[ -n $LSST_CONDA_CHANNELS ]]; then
+      # remove any previously configured non-default channels
+      # XXX allowed to fail
+      conda config --env --remove-key channels || true
```
ktlim, line 304: Add `2> /dev/null` to prevent scary errors from getting to the user.

PR #263, opened 2020-08-04
Added check on existing environment

gcomoretto:


brianv0:
looks fine. I think this might be doing an unnecessary conda clean, but that operation will be quick, and I think it's fine.

PR #264, opened 2020-08-05
Updated save_build_env to use --explicit and improved

gcomoretto:
I have also improved the logic and the output a little bit.

### DIFF
--- a/bin/rebuild
+++ b/bin/rebuild
@@ -37,37 +37,51 @@ save_build_env() {
   mkdir -p "${LSSTSW_BUILD_DIR}/builds"
   filename="${LSSTSW_BUILD_DIR}/builds/${BUILD}.env"
   echo "Saving environment information in ${filename}"
-  tmp_file="$(mktemp)"
+
   active_env="$(mktemp)"
   ref_env="$(mktemp)"
-  conda env export > "${tmp_file}"
+
+  # saving information in envinronment name
   echo "environment_name: ${LSST_CONDA_ENV_NAME}" > "${filename}"
-  grep "^prefix" "${tmp_file}" >> "${filename}"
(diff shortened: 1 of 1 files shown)


bin/rebuild
```diff
@@ -37,37 +37,53 @@ save_build_env() {
...
+    } >> "${filename}"
+    grep -v "^name" "${local_env_file}" > "${ref_env}"
+    # save diff in file
+    echo "Differences between active anvironment and reference environment:" >> "${filename}"
+    echo "Differences between active anvironment and reference environment:"
+    set +e  # disable set -e for the diff commands.
+    diff -B "${active_env}" "${ref_env}" >> "${filename}"
+    # show diff in output
+    diff -u "${active_env}" "${ref_env}"
+    set -e
```
ktlim, line 81: `set +e/-e` is a bit dangerous, as it is global state. Better to remove them and instead protect the diffs.
```suggestion
    diff -B "${active_env}" "${ref_env}" >> "${filename}" || true  # OK for this to fail
    # show diff in output
    diff -u "${active_env}" "${ref_env}" || true  # OK for this to fail
```

bin/rebuild
```diff
@@ -37,37 +37,53 @@ save_build_env() {
   mkdir -p "${LSSTSW_BUILD_DIR}/builds"
   filename="${LSSTSW_BUILD_DIR}/builds/${BUILD}.env"
   echo "Saving environment information in ${filename}"
-  tmp_file="$(mktemp)"
+
   active_env="$(mktemp)"
   ref_env="$(mktemp)"
-  conda env export > "${tmp_file}"
+
+  # saving information in envinronment name
```
ktlim, line 44:
```suggestion
  # saving information in environment name
```

PR #268, opened 2020-09-18
Avoid  environment consistency check when executing bin/deploy

gcomoretto:
I have added an extra option (-R) to check the consistency of the environment.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -45,7 +45,7 @@ usage() {
 
 		Specific options:
 	    -b          use bleeding edge conda packages
-	    -r REF      git ref in scipipe_conda_env (hash, branch, tag), create new end only if it does not exists
+	    -r REF      git ref in scipipe_conda_env (hash, branch, tag), create new env only if it does not exists
 	    -R REF      like -r, but always check if it is consistent with the pinned file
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -261,53 +268,57 @@ main() {
...
-    # disable the conda install progress bar when not attached to a tty. Eg.,
-    # when running under CI
-    if [[ ! -t 1 ]]; then
-      ARGS+=("--quiet")
-    fi
+      # disable the conda install progress bar when not attached to a tty. Eg.,
+      # when running under CI
+      if [[ ! -t 1 ]]; then
+        ARGS+=("--quiet")
+      fi
```
brianv0, line 315: Is this tabs vs spaces (and correct?)

brianv0:
Looks fine, just a question of tabs/spaces are correct (it's hard to tell from github interface)

bin/deploy
gcomoretto: I don't think there should be tabs here. I just indented 2 spaces more, to keep (visual) consistency with the rest, but I assume it makes no difference.

ktlim (approved):

PR #269, opened 2020-09-22
DM-26170: Migrate ci_cpp from lsst-dm to lsst

czwa:
Register testdata_latiss_cpp.


timj (approved):

timj:
Please can you add testdata_ci_hsc here as well since it's missing and seems like it should be there for completeness.

PR #270, opened 2020-09-30
Improving log and resolving a couple of errors

gcomoretto:
- made the messages consistent
- moved the --quite option where it can be used
- conda clean not done when the environment exists and it is consistent
- sorted the lock files when compared
- removed the subshell (I should reduce the indentation, but that would make the review more difficult, I'll do before merging)

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -326,7 +326,7 @@ main() {
   # shellcheck disable=SC1091
   conda activate "${LSST_CONDA_ENV_NAME}"
 
-  # Configuring propper channles at environment level
+  # Configuring proper channels at environment level
   (
     # configure alt conda channel(s)
     if [[ -n $LSST_CONDA_CHANNELS ]]; then
--- a/bin/deploy
+++ b/bin/deploy
@@ -271,22 +271,24 @@ main() {
   if conda env list | grep "^$LSST_CONDA_ENV_NAME " > /dev/null && [[ $CHECKENV = false ]]; then
     echo "::: Environment $LSST_CONDA_ENV_NAME already deployed."
   else
-    (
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -295,37 +297,36 @@ main() {
         if (conda run -n "${LSST_CONDA_ENV_NAME}" conda list --explicit > "${tmp_lock}"); then
```
ktlim, line 297: Does this need a `2>/dev/null` in case there is no such environment?

bin/deploy
```diff
@@ -295,37 +297,36 @@ main() {
...
-    )
   fi
 
   # intentionally outside of a lockfile subshell
-  # shellcheck disable=SC1091
   echo "Activating environment ${LSST_CONDA_ENV_NAME}"
   # shellcheck disable=SC1091
   conda activate "${LSST_CONDA_ENV_NAME}"
 
+  # Configuring propper channles at environment level
```
ktlim, line 329:
```suggestion
  # Configuring proper channels at environment level
```

bin/deploy
```diff
@@ -339,7 +340,7 @@ main() {
 
       conda config --env --set channel_priority strict
 
-      conda config --show
+      #conda config --show
```
ktlim, line 343: Normally we don't leave commented-out debug code.

bin/deploy
```diff
@@ -349,7 +350,7 @@ main() {
   # canoncial env files, doesn't seem to list pip installed packages.  The the
   # "newer" yaml env file format from `conda env export` does include pip/pypi
   # packages.
-  conda env export
+  #conda env export
```
ktlim, line 353: Again, remove debug code and maybe the preceding comment as well.

ktlim (approved):

PR #279, opened 2021-01-27
DM-28473: Use zsh-friendly LSSTSW computation.

ktlim:


timj (approved):
Looks okay. In python I'd like to refactor to put the zsh test set a new variable that is either `$0` or `${BASH_SOURCE[0]}` since that would prevent the duplication of the `dirname` logic but it's not really worth it.

It does seem odd that we are checking for "not zsh" rather than "is bash" when deciding to use a bash-specific variable but I take it that "is bash" is harder to know?

ktlim:
Actually I was just copying what had been there before.  It is possible to know if it is bash.  I'll refactor before merging.

timj:
Yeah, I just saw the exact same code in the envconfig script.

PR #288, opened 2021-06-17
DM-30792: Enable deployment of exact eups-tagged environments

ktlim:


bin/deploy
```diff
@@ -40,6 +40,7 @@ usage() {
 	    -v VERSION  a specific version of rubin-env conda package.
 	                By default will use the version specified in etc/settings.cfg.sh
 	    -r REF      git reference (SHA1) in scipipe_conda_env repository (backward compatibility)
+            -x TAG      eups tag to select the exact environment for that tag
```
timj, line 43: There seems to be a tab/space issue here.

I also think "exact" is a bit strong isn't it? Aren't we deriving the rubin-env but then installing whatever conda gives us for that version? We aren't installing the exact environment that was used to build the eups tag.

bin/deploy
```diff
@@ -171,13 +200,21 @@ main() {
       for c in $LSST_CONDA_CHANNELS; do
         ARGS+=("-c $c")
       done
-      ARGS+=("rubin-env=${RUBIN_ENV_VERSION}")
+      if [[ -n $TAG ]]; then
+	ARGS+=("--file" "${TAG}.env")
```
timj, line 204: indent?

bin/deploy
```diff
@@ -90,6 +95,27 @@ am_I_sourced() {
   fi
 }
 
+get_tagged_env() {
+  eups_root=https://eups.lsst.codes/stack
+  env_version=$($CURL "${CURL_OPTS[@]}" -L "$eups_root/src/tags/$1.list" \
```
timj, line 100: I tried to take a look at https://eups.lsst.codes/stack/src/tags and it seems to be broken because clicking on any of the links takes you to a non-functioning link of things like https://eups.lsst.codes/stack/src/w_2021_25.list rather than the expected https://eups.lsst.codes/stack/src/tags/w_2021_25.list

bin/deploy
```diff
@@ -40,6 +40,7 @@ usage() {
 	    -v VERSION  a specific version of rubin-env conda package.
```
timj, line 40: Maybe we should complain if `-v` and `-x` are given?

timj (approved):
Minor comments.

bin/deploy
ktlim: No, this is installing the exact environment (via `--file` with `@EXPLICIT`).

bin/deploy
ktlim: You need a trailing slash when viewing in the browser.

PR #289, opened 2021-07-02
DM-30981: Use rubin-env version for eups stack path.

ktlim:


timj (approved):
Looks okay. I had not heard of `jq` before so thanks for that.

PR #291, opened 2021-08-05
DM-31303: Ensure channel priority.

ktlim:


timj (approved):
This seems consistent with what was done in lsst/lsst#145

PR #296, opened 2022-01-23
DM-33354: Use mamba to create environment.

ktlim:


bin/deploy
```diff
@@ -204,16 +205,16 @@ main() {
       fi
       ARGS+=('--name' "$LSST_CONDA_ENV_NAME")
       for c in $LSST_CONDA_CHANNELS; do
-        ARGS+=("-c $c")
+        ARGS+=("-c$c")
```
erykoff, line 208: Does `—channel conda-forge` work? Would look better at least.

erykoff (approved):
Should work but a comment.

bin/deploy
erykoff: That should be two dashes. Stupid autocorrect.

ktlim: It does not.

PR #297, opened 2022-01-23
DM-33354 (hotfix): Always activate base package (esp. before installing mamba).

ktlim:


bin/deploy
```diff
@@ -205,7 +206,7 @@ main() {
       fi
       ARGS+=('--name' "$LSST_CONDA_ENV_NAME")
       for c in $LSST_CONDA_CHANNELS; do
-        ARGS+=("-c$c")
+        ARGS+=("-c" "$c")
```
erykoff, line 209: Ah, I see what the problem was before.  Unsure how it ever worked though!

erykoff (approved):
Looks good, and what we should be doing.  Don't know if it will fix the underlying issue, of course.

PR #301, opened 2022-03-23
DM-34165: Simplify environment menu generation.

ktlim:


erykoff (approved):
If it works, it works, but I'm not an expert bash programmer.  (And is everything here pure posix?  Seems so...)

PR #307, opened 2023-01-31
DM-4131: Add -c option to tag as current.

ktlim:


ktlim:
A doc-only fix was preferred.

PR #308, opened 2023-02-01
DM-37817: Replace bespoke script with eups subcommand.

ktlim:


timj (approved):
Nice cleanup to now use eups native feature.

PR #310, opened 2023-04-19
Improve help strings for -p/-t docs

parejkoj:
Related to [DM-4131](https://jira.lsstcorp.org/browse/DM-4131): I just did a build and completely forgot about the `-t current` option that KT had pointed out on that ticket, because when I ran `-h` to get help, I didn't understand what the help message meant.

I also tweaked the help message for `-p`, as the meaning of "prepare" wasn't obvious.


ktlim (approved):
LGTM.

PR #311, opened 2023-06-08
DM-39595: Always use latest miniconda.

ktlim:


timj (approved):
Maybe add a comment at the top stating why that env var is still set.

timj (approved):
Looks great and testing on arm shows that it works for me without @erykoff 's branch.

PR #312, opened 2023-06-21
DM-39738: Add nodefaults channel to prevent use of defaults.

ktlim:


bin/deploy
```diff
@@ -197,7 +197,7 @@ main() {
       # installation attempt.  These need to be cleaned up before [re]attempting
       # to install packages.
       conda clean -y --all
-      conda install -y -c conda-forge mamba
+      conda install -y -c conda-forge -c nodefaults mamba
```
timj, line 200: Does it matter that this is a different order to that in LSST_CONDA_CHANNELS?

timj (approved):

bin/deploy
ktlim: Nope.

PR #318, opened 2023-10-11
DM-41137: Remove outdated build products

JeremyMcCormick:
The qserv_distrib and dax_webserv products are outdated and do not build any longer. I have removed them from the products list so that only lsst_distrib will be built by default now when calling rebuild with no arguments.


timj (approved):
Looks good.

PR #319, opened 2023-10-11
DM-41137: Remove quotes when appending products to args array

JeremyMcCormick:
This fixes a problem where products defined in the $PRODUCTS variable were incorrectly interpreted as a single string, causing the build to fail. The shellcheck tool complains about this syntax but it is the most straightforward way to perform this operation in bash.


JeremyMcCormick:
I'm going to close this for now and reopen later when I have found a good solution. The changes I have on this branch are not working correctly right now.

PR #320, opened 2023-10-11
DM-41151: Define products variable as array so that it is interpreted correctly

JeremyMcCormick:
The usage of quotes causes the products to be interpreted as a single string, which is incorrect. This breaks the rebuild command when multiple products are included in the `$PRODUCTS` variable.

### DIFF
--- a/etc/settings.cfg.sh
+++ b/etc/settings.cfg.sh
@@ -11,7 +11,7 @@ SPLENV_BASE_NAME="lsst-scipipe"
 LSST_SPLENV_REPO=${LSST_SPLENV_REPO:-https://github.com/lsst/scipipe_conda_env.git}
 
 # top-level products
-PRODUCTS='lsst_distrib'
+PRODUCTS=("lsst_distrib")
 
 # change to ssh+git if push is needed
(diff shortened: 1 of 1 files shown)


JeremyMcCormick:
@ktlim This change fixes handling of multiple arguments defined in `$PRODUCTS` but breaks the shellcheck GH action, as it wants all variables to be double-quoted. This doesn't work when appending values to a bash array.

Is there an alternate method of doing this that would be preferable and also keep shellcheck happy?

JeremyMcCormick:
I've added a shellcheck directive to allow usage of an unquoted variable for this one line. 

This seems reasonable since arrays shouldn't be quoted when appending to another array in bash, as it needs to split on spaces rather than interpret the variable as a single string.

ktlim (approved):

JeremyMcCormick:
@ktlim suggested changing the PRODUCTS variable itself to an array, which fixed the issue. I rebased and discarded the other changes to keep that single commit. 

I've checked that `rebuild` and `rebuild lsst_distrib lsst_ci` work properly. I also put multiple products into the array for testing, and that worked as well (checks done on S3DF at SLAC).

PR #325, opened 2024-08-09
DM-45687: Add a timer to rebuild script

roceb:


bin/rebuild
```diff
@@ -217,8 +218,9 @@ fi
   if [[ -n $TAG ]]; then
     eups tags --clone="$BUILD" "$TAG"
   fi
-
-  echo "# BUILD $BUILD completed."
+  timer_end=$(date +%s)
+  RUNTIME=$((timer_end-timer_start))
+  echo "# BUILD $BUILD completed. Took $RUNTIME seconds to complete"
```
timj, line 223:
```suggestion
  echo "# BUILD $BUILD completed in $RUNTIME seconds."
```
I think I prefer not to say "complete" twice.

timj (approved):
Looks good. Minor tweaking of sentence requested.

bin/envconfig
```diff
@@ -72,15 +72,6 @@ if [ "$finish" = true ]; then
   return
 fi
 
-if [[ -n $BASH_VERSION ]]; then
```
timj, line 75: What's this change about? It's not part of the stated goal of this commit.

roceb: Committed by mistake.

PR #327, opened 2025-01-19
DM-46554: Update eups platform address for arm64

roceb:


ktlim (approved):
I guess I was kind of hoping we could start getting rid of the `miniconda-` part as part of this transition, but that can come later.

PR #334, opened 2025-03-16
DM-49474: Add flag to install a specific rubin-env

mwittgen:


bin/deploy
```diff
@@ -34,13 +34,14 @@ usage() {
...
 
-		Usage: $0 [-h] [-r REF|-v VERSION|-x TAG]
+		Usage: $0 [-h] [-r REF|-v VERSION|-x TAG] [-R ENV]
 
 		Specific options:
 	    -v VERSION  a specific version of rubin-env conda package.
 	                By default will use the version specified in etc/settings.cfg.sh
 	    -r REF      git reference (SHA1) in scipipe_conda_env repository (backward compatibility)
 	    -x TAG      eups tag to select the exact environment for that tag
+    -R ENV      install a specific rubin-env, like rubin-env-developer
```
ktlim, line 44:
```suggestion
    -R ENV      install a specific rubin-env variant, like rubin-env-developer
```

bin/deploy
```diff
@@ -34,13 +34,14 @@ usage() {
   # note that heredocs are prefixed with tab chars
   fail "$(cat <<-EOF
 
-		Usage: $0 [-h] [-r REF|-v VERSION|-x TAG]
+		Usage: $0 [-h] [-r REF|-v VERSION|-x TAG] [-R ENV]
```
ktlim, line 37:
```suggestion
		Usage: $0 [-h] [-r REF|-v VERSION|-x TAG] [-R VARIANT]
```

bin/deploy
```diff
@@ -225,7 +229,7 @@ main() {
       if [[ -n $TAG ]]; then
         ARGS+=("--file" "${TAG}.env")
       else
-        ARGS+=("rubin-env=${RUBIN_ENV_VERSION}")
+        ARGS+=("${LSST_RUBIN_ENV_NAME}=${RUBIN_ENV_VERSION}")
```
ktlim, line 232: Since bin/deploy is for developers anyway, I guess it's OK that this now allows installation of any conda package instead of rubin-env.

ktlim (approved):

bin/deploy
mwittgen: Added some additional checks

roceb: My only comment is whether -R is a good flag or not. Otherwise this looks good to me

mwittgen: -V instead?

PR #335, opened 2025-03-18
DM-49530: Fix check for osx arm

mwittgen:


roceb (approved):
Don't know where I got the 64 from. Good catch

PR #338, opened 2025-05-07
DM-50758: Added posthook script to fix shebang on binaries 

roceb:


bin/envconfig
```diff
@@ -161,6 +185,7 @@ if [ ! -d "$EUPS_PATH" ]; then
   echo "::: Deploying manifest.remap"
   ln -sf "${LSSTSW}/etc/manifest.remap" "${EUPS_PATH}/site/manifest.remap"
 fi
+copy_postinstall_fix
```
timj, line 188: Please put this inside the above `if` block. It seems to me that the code creating the directory and putting in the manifest remap file should be the same code that is copying in the post install hook. I don't expect a situation where someone is creating the directory themselves and then getting confused when binary installs break (they won't have the manifest remap either so will have other problems). If we do that we can simplify the other code that is checkign for the presence of the hook and telling us it is checking for the hook.

timj (approved):
Looks great. Thanks. Can you please turn this into two commits: 1 to clean up the unnecessary EUPS_PATH setting in deploy, and the other adding the hook installation?

